PR #799, opened 2025-10-13
feat(APP-4616): Implement proposal execute guard

milosh86:
# Description

Implement proposal execute guard by calling `dao.hasPermission()` for plugins with restricted execute.

Task: [APP-4616](https://aragonassociation.atlassian.net/browse/APP-4616)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Developer Checklist:

- [x] Manually smoke tested the functionality locally

## Review Checklist:

- [ ] Tested locally that all Acceptance Criteria or Expected Outcomes are satisfied

[APP-4616]: https://aragonassociation.atlassian.net/browse/APP-4616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/plugins/multisigPlugin/utils/multisigPluginUtils/multisigPluginUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+            release: plugin.release,
+        });
+        const hasExecuteProposalPermissionGuard = versionComparatorUtils.isGreaterOrEqualTo(normalisedPluginVersion, {
+            release: 1,
+            build: 3,
+        });
+
+        return hasExecuteProposalPermissionGuard;
+    };
+}
```
thekidnamedkd, line 21: Seems like we could flatten/centralize this util as shared for multisig/token/etc as it's the same logic and even the same version number... I'm also wondering if we should pin the release and build numbers somewhere more shared/isolated as we don't expect this to change often, it's more about supporting backwards compatibility of a specific function than version checking? Unless you think this could be generalized for more use cases such as other contract functions. That would support the argument for making it shared more so.

thekidnamedkd (commented):
Looking good. I left one sizeable comment. I didn't see a design for this to review against, but I trust it's pretty standard?

src/plugins/multisigPlugin/utils/multisigPluginUtils/multisigPluginUtils.ts
milosh86: > Seems like we could flatten/centralize this util as shared for multisig/token/etc as it's the same logic and even the same version number... 

I was thinking about this and was not sure since the whole logic is around using `versionComparatorUtils`, but maybe I could create some shared util like `isPluginVersionGreaterOrEqualTo`.

I don't expect this to change too frequently (if at all). If it does, than I would rather look for a better solution than to "scale" this one. 

> I'm also wondering if we should pin the release and build numbers somewhere more shared/isolated

Not sure what would we get with putting this in a central place, i.e. plugin info? This function is a plugin related logic to figure out if `execute` is restricted or not, and it happens that the logic is based on versions, but we might change this.

src/shared/utils/pluginMetaUtils/pluginMetaUtils.ts
```diff
@@ -0,0 +1,16 @@
+import type { IDaoPlugin } from '../../api/daoService';
```
thekidnamedkd, line 1: This is extremely nit-picky, but does it makes sense to call this pluginMetaUtils or something more broad as a shared set of methods that may grow, and then the method becomes 'isVersionGreaterOrEqualTo' 

or...

Just add to pluginTransactionUtils or permissionTransactionUtils?

thekidnamedkd (commented):
Niceee. See my annoying comment 😅

src/shared/utils/pluginMetaUtils/pluginMetaUtils.ts
milosh86: I was thinking about that...but got into analysis paralysis and just created this util :D ... I like `pluginMetaUtils` 👍

milosh86: let me refactor quickly

thekidnamedkd (approved):
killer ✨

PR #887, opened 2025-12-01
fix(APP-372): Fix DAO display without plugins

tyhonchik:
# Description

The `/v2/plugins/by-dao/:network/:address/details` endpoint was returning 404 while being called as part of DAO loading, which surfaced as a parsing/backend error and caused DAO pages to fail instead of simply treating that situation as “this DAO has no plugins.”

- Improves `AragonBackendServiceError` to handle bare 404 responses from the backend.
- Updates error handling logic.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/.changeset/sharp-sheep-appear.md
+++ b/.changeset/sharp-sheep-appear.md
@@ -1,0 +1,5 @@
+---
+'@aragon/app': patch
+---
+
+Improve handling for DAOs which have no installed plugins
--- a/src/shared/api/aragonBackendService/aragonBackendServiceError.ts
+++ b/src/shared/api/aragonBackendService/aragonBackendServiceError.ts
@@ -20,25 +20,25 @@ export class AragonBackendServiceError extends Error {
     }
 
     static fromResponse = async (response: Response): Promise<AragonBackendServiceError> => {
-        try {
-            const parsedData = await responseUtils.safeJsonParse(response);
+        const parsedData = await responseUtils.safeJsonParse(response);
 
-            const isIErrorResponse = (value: unknown): value is IErrorResponse =>
-                value != null &&
-                typeof value === 'object' &&
-                'code' in (value as Record<string, unknown>) &&
-                typeof (value as Record<string, unknown>).code === 'string' &&
-                'description' in (value as Record<string, unknown>) &&
-                typeof (value as Record<string, unknown>).description === 'string';
+        const isIErrorResponse = (value: unknown): value is IErrorResponse =>
+            value != null &&
+            typeof value === 'object' &&
+            'code' in (value as Record<string, unknown>) &&
+            typeof (value as Record<string, unknown>).code === 'string' &&
+            'description' in (value as Record<string, unknown>) &&
+            typeof (value as Record<string, unknown>).description === 'string';
 
-            if (isIErrorResponse(parsedData)) {
-                return new AragonBackendServiceError(parsedData.code, parsedData.description, response.status);
-            }
-
-            return new AragonBackendServiceError(this.parseErrorCode, this.parseErrorDescription, response.status);
-        } catch {
-            return new AragonBackendServiceError(this.parseErrorCode, this.parseErrorDescription, response.status);
+        if (isIErrorResponse(parsedData)) {
+            return new AragonBackendServiceError(parsedData.code, parsedData.description, response.status);
         }
+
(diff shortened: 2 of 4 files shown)


tyhonchik:
test on `/dao/ethereum-mainnet/yb.dao.eth/dashboard`

.changeset/sharp-sheep-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@aragon/app': patch
+---
+
+Fix DAOs without plugins
```
thekidnamedkd:
```suggestion
Improve handling for DAOs which have no installed plugins
```

src/shared/api/aragonBackendService/aragonBackendServiceError.ts
```diff
@@ -32,11 +36,30 @@ export class AragonBackendServiceError extends Error {
...
+                    response.status,
+                );
+            }
+
             return new AragonBackendServiceError(this.parseErrorCode, this.parseErrorDescription, response.status);
         } catch {
+            if (response.status === 404) {
+                return new AragonBackendServiceError(
+                    this.httpNotFoundCode,
+                    `Backend endpoint returned 404: ${response.url}`,
```
thekidnamedkd: Can't remember the pattern specifically but I think the descriptions should be translated if custom?

src/shared/api/aragonBackendService/aragonBackendServiceError.ts
```diff
@@ -32,11 +36,30 @@ export class AragonBackendServiceError extends Error {
...
             if (isIErrorResponse(parsedData)) {
+                // Domain-level error coming from backend (e.g. { code: "notFound", ... }).
                 return new AragonBackendServiceError(parsedData.code, parsedData.description, response.status);
             }
 
+            // Backend responded with 404 but without our standard error shape:
+            // treat it as "endpoint missing" instead of a domain not-found.
+            if (response.status === 404) {
+                return new AragonBackendServiceError(
+                    this.httpNotFoundCode,
```
thekidnamedkd: Does the Not Found code make sense for this then? I believe technically it's found but just does not contain the field we try to spread onto a new object... I think that this error handling is maybe a good improvement, but we should also actually handle this case with the construction of the { ...dao, plugins: [] } object which is actually correct and should be supported if a DAO has uninstalled all plugins.

src/shared/api/daoService/daoService.ts
```diff
@@ -40,10 +41,15 @@ class DaoService extends AragonBackendService {
...
+        try {
+            const { network, address } = this.parseDaoId(dao.id);
+            const plugins = await pluginsService.getPluginsByDao({ urlParams: { network, address } });
 
-        return { ...dao, plugins };
+            return { ...dao, plugins };
+        } catch (error) {
+            monitoringUtils.logError(error);
+            return { ...dao, plugins: [] };
+        }
```
thekidnamedkd, line 52: Not sure if this is backend improvement but we should probably be able to handle no plugins before the catch block as this is potentially not an error case, definitely an edge case however. Let's define the paths from Evan, I can't remember the specific reqs of this DAO impl.

thekidnamedkd (commented):
Left a few comments and questions. Let's clarify what we expect from this case, it's definitely funky 😅

.changeset/sharp-sheep-appear.md
tyhonchik: Done
Thanks

src/shared/api/aragonBackendService/aragonBackendServiceError.ts
tyhonchik: Removed text
Just made error message more helpful

src/shared/api/aragonBackendService/aragonBackendServiceError.ts
tyhonchik: Not found meant "endpoint doesn't exist" in this part of code


anyway I simplified the AragonBackendServiceError

src/shared/api/daoService/daoService.ts
tyhonchik: here I wrap usage of getting plugins from separate endpoint

if there any problem I return dao itself

thekidnamedkd (approved):
LGTM ✨

PR #917, opened 2025-12-17
fix(APP-390, APP-391): Allow wizard to progress with continue button on simulate; Fix hydration mismatches and navigation blocking side effects

tyhonchik:
# Description

### APP-390
Fixes a bug where the simulate actions dialog was closing all dialogs (including the publish proposal dialog) when continuing from the final wizard step.

**Root cause:** The dialog called `close()` without an ID, which closes all dialogs in the stack.

**Fix:** Changed to `close(location.id)` to only close the simulate dialog itself. Removed the `setTimeout` workaround and added regression tests.

### APP-391
Fixes hydration mismatches and navigation blocking side effects in transaction dialogs and wallet UI.

#### Changes

- **Hydration fix (Wallet/nav)**: Defer `useAccount()`-derived rendering until after mount to prevent SSR/client mismatch when wallet auto-reconnects
- **React warning fix**: Remove `setIsBlocked(false)` from `href` builder functions to avoid "setState during render" errors
- **Navigation unblock**: Unblock navigation when success/proceed-anyway state is reached (before user clicks) to prevent unwanted confirm dialogs
- **Link guard fix**: Move navigation blocking check to capture phase and stop propagation on cancel to prevent Next.js top loader from starting

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/src/modules/governance/dialogs/simulateActionsDialog/simulateActionsDialog.tsx
+++ b/src/modules/governance/dialogs/simulateActionsDialog/simulateActionsDialog.tsx
@@ -4,6 +4,7 @@ import { useTranslations } from '@/shared/components/translationsProvider';
 import { ActionSimulation, Dialog, invariant } from '@aragon/gov-ui-kit';
 import { useEffect } from 'react';
 import { useSimulateActions } from '../../api/actionSimulationService';
+import { GovernanceDialogId } from '../../constants/governanceDialogId';
 import type { IProposalCreateAction } from '../publishProposalDialog';
 
 export interface ISimulateActionsDialogParams {
@@ -35,7 +36,6 @@ export const SimulateActionsDialog: React.FC<ISimulateActionsDialogProps> = (pro
 
     const { t } = useTranslations();
     const { close } = useDialogContext();
-    const dialogId = location.id;
 
     const { mutate: simulateActions, isError, isPending, status, data } = useSimulateActions();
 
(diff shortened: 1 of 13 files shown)


src/modules/governance/dialogs/simulateActionsDialog/simulateActionsDialog.tsx
```diff
@@ -55,19 +56,20 @@ export const SimulateActionsDialog: React.FC<ISimulateActionsDialogProps> = (pro
...
+        // Important: close only this dialog (by id) to avoid accidentally closing dialogs opened by the submit handler.
+        if (formId != null) {
             const form = document.getElementById(formId);
             if (form instanceof HTMLFormElement) {
                 form.requestSubmit();
-                // Defer close to allow form submission to process
-                setTimeout(() => close(), 0);
             }
         }
+        close(dialogId);
```
milosh86: we should probably use simulationActionDialog ID explicitly here (like in other similar cases)

milosh86 (commented):
looks good, left minor comment.

Also, the following logic is used only in 2 places for now, so it's ok, but if we start adding it to let's say 2 more, then let's create a hook for it.

```
 const isMounted = useIsMounted();
 const effectiveIsConnected = isMounted && isConnected;
```

I'll check preview now.

src/modules/governance/dialogs/simulateActionsDialog/simulateActionsDialog.tsx
tyhonchik: makes sense
I'll change it
anyway `location.id` should be equal simulationActionDialog ID

milosh86 (approved):
All good, checked the preview 👍

PR #1006, opened 2026-03-02
fix(APP-531): add safe BigInt parsing to prevent runtime crashes

tyhonchik:
# Description

- Replace all raw BigInt() calls that take external API data with bigIntUtils.safeParse()
- Fixes crash where totalSupply arrived as "10000000000000000000000000.0" from causing "Cannot convert to a BigInt" runtime error.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/src/shared/utils/bigIntUtils/bigIntUtils.ts
+++ b/src/shared/utils/bigIntUtils/bigIntUtils.ts
@@ -1,4 +1,8 @@
 class BigIntUtils {
+    private readonly scientificNotationRegex =
+        /^([+-]?)(\d+\.?\d*)[eE]([+-]?\d+)$/;
+    private readonly hexNotationRegex = /^0[xX][0-9a-fA-F]+$/;
+
     /**
      * Safely converts a value to BigInt, handling floating-point representations
      * (e.g. "10000000000000000000000000.0") and scientific notation (e.g. "1e+25")
@@ -30,8 +34,12 @@ class BigIntUtils {
             return fallback;
         }
 
-        if (/[eE]/.test(str)) {
-            return this.parseScientificString(str) ?? fallback;
+        const scientificValue = this.parseScientificString(str);
+        if (scientificValue != null) {
+            return scientificValue;
+        }
+        if (/[eE]/.test(str) && !this.hexNotationRegex.test(str)) {
+            return fallback;
         }
 
         const integerPart = str.split('.')[0];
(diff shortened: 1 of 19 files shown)


src/shared/utils/bigIntUtils/bigIntUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            // Fractional result — truncate towards zero
+            const divisor = BigInt(10) ** BigInt(-adjustedExp);
+            return significand / divisor;
+        } catch {
+            return null;
+        }
+    };
+}
```
thekidnamedkd, line 87: Need to handle hex values with e/E (like 0x1e) more safely. Refactored this into a reusable helper path so we don’t mis-detect them as scientific notation.

```suggestion
 safeParse = (
        value: string | number | bigint | null | undefined,
        fallback: bigint = BigInt(0),
    ): bigint => {
        if (value == null) {
            return fallback;
        }
        if (typeof value === 'bigint') {
            return value;
        }

        if (typeof value === 'number') {
            if (!Number.isFinite(value)) {
                return fallback;
            }
            return BigInt(Math.trunc(value));
        }

        const str = value.trim();
        if (str === '') {
            return fallback;
        }

        const scientificValue = this.parseScientificString(str);
        if (scientificValue != null) {
            return scientificValue;
        }

        const integerPart = str.split('.')[0];

        try {
            return BigInt(integerPart);
        } catch {
            return fallback;
        }
    };

    /**
     * Parses a scientific notation string (e.g. "1.5e+25", "-3e10") directly into BigInt
     * without going through Number, preserving full precision for the significant digits.
     */
    private parseScientificString = (str: string): bigint | null => {
        const match = str.match(this.scientificNotationRegex);
        if (!match) {
            return null;
        }

        const [, sign, mantissa, expStr] = match;
        const exp = Number(expStr);

        const dotIndex = mantissa.indexOf('.');
        const fracLength = dotIndex >= 0 ? mantissa.length - dotIndex - 1 : 0;
        const digits = mantissa.replace('.', '');

        const adjustedExp = exp - fracLength;

        try {
            const significand = BigInt((sign === '-' ? '-' : '') + digits);

            if (adjustedExp >= 0) {
                return significand * BigInt(10) ** BigInt(adjustedExp);
            }

            // Fractional result — truncate towards zero
            const divisor = BigInt(10) ** BigInt(-adjustedExp);
            return significand / divisor;
        } catch {
            return null;
        }
```

thekidnamedkd (commented):
Looks good! Just left one comment for an edge case, but seems like it's pretty plausible.

src/shared/utils/bigIntUtils/bigIntUtils.ts
tyhonchik: good catch, thanks
fixed

thekidnamedkd (approved):

PR #1027, opened 2026-03-13
feat(APP-551): Add member list onboarding entry point cards

milosh86:
# Description

Adds onboarding entry point cards to the token member list. When the connected user has not yet delegated their tokens or has not yet locked/wrapped them, the relevant onboarding card is surfaced inline in the member list:

- **Delegation card** — shown when the user has a token balance but no delegate set
- **Lock card** — shown when the plugin has a voting escrow and the user has underlying tokens but no votes
- **Wrap card** — shown when the plugin has no voting escrow and the user has underlying tokens but no wrapped balance

Delegation card takes priority over lock/wrap card when both conditions are true.

Also updates the delegation onboarding dialog entry point to skip the intro step.

Related to: APP-551

## Type of Change

- [x] Minor: Feature (non-breaking change which adds new functionality)

## Developer Checklist:

- [ ] Manually smoke tested the functionality in a preview or locally
- [ ] Confirmed there are no new warnings or errors in the browser console
- [ ] (For User Stories only) Double-checked that all Acceptance Criteria are satisfied
- [ ] Confirmed there are no new warnings on automated tests
- [ ] Merged and published any dependent changes in downstream modules
- [ ] Selected the correct base branch
- [ ] Commented the code in hard-to-understand areas
- [ ] Followed the code style guidelines of this project
- [ ] Reviewed that the Files Changed in Github's UI reflect my intended changes
- [ ] Confirmed the pipeline checks are not failing

## Review Checklist:

- [ ] (For User Stories only) Tested in a preview or locally that all Acceptance Criteria are satisfied
- [ ] Confirmed that changes follow the code style guidelines of this project

PR #1039, opened 2026-03-18
feat(APP-553): Pin “You” and “Your delegate” tags on token members list

thekidnamedkd:
### Description

  Implements APP-553: Show "You" and "Your delegate" tags on the Members page for token-voting DAOs.

### Changes:
- TokenMemberList: Added delegate lookup via useTokenCurrentDelegate and separate useMember fetches for the connected user and their
delegate. This guarantees pinning to the top of the list regardless of pagination position. Pinned entries are deduplicated against the
 paginated list.
- TokenMemberListItem: Added isDelegate prop, passed through to MemberDataListItem.Structure from gov-ui-kit which renders the "Your
Delegate" tag.
- "You" tag: Handled automatically by gov-ui-kit's MemberDataListItem.Structure via internal useAccount() detection — no code needed.
- Pinning logic: Connected user pinned at index 0 (if voting power > 0), delegate pinned at index 1 (if delegate is not zero address
and not self). Uses useMemo for the merged list.
- Tests: Updated existing tests and added 7 new test cases covering pinning, delegation edge cases (zero address, self-delegation),
deduplication, and voting power conditions.

A user can have both tags simultaneously — if they delegate their own tokens away but receive delegations from others, their VP > 0
(from received delegations) and they have a valid delegate. "You" is always pinned first in this scenario.

  Affects both tokenPlugin and lockToVotePlugin member lists (shared component).

  Closes APP-553

### Type of Change

### Developer Checklist:

### Review Checklist:


src/plugins/tokenPlugin/components/tokenMemberList/tokenMemberList.tsx
```diff
@@ -1,14 +1,24 @@
...
 import {
+    addressUtils,
     DataListContainer,
     DataListPagination,
     DataListRoot,
     MemberDataListItem,
 } from '@aragon/gov-ui-kit';
+import { useMemo } from 'react';
+import { zeroAddress } from 'viem';
+import { useAccount } from 'wagmi';
```
tyhonchik: useaccount should be deprecated in wagmi 3
use `useConnection` instead

src/plugins/tokenPlugin/components/tokenMemberList/tokenMemberList.test.tsx
```diff
@@ -23,6 +35,14 @@ describe('<TokenMemberList /> component', () => {
         'useMemberListData',
     );
     const useDaoSpy = jest.spyOn(daoService, 'useDao');
+    const useAccountSpy = jest.spyOn(wagmi, 'useAccount');
```
tyhonchik: here as well should be useConnection

src/plugins/tokenPlugin/components/tokenMemberList/tokenMemberList.tsx
```diff
@@ -29,7 +40,86 @@ export const TokenMemberList: React.FC<ITokenMemberListProps> = (props) => {
...
+
+        if (connectedUserMember?.address) {
+            const paginatedEntry = memberList.find((m) =>
+                addressUtils.isAddressEqual(
+                    m.address,
+                    connectedUserMember.address,
+                ),
+            );
+            const entry = paginatedEntry ?? connectedUserMember;
+            const hasVotingPower = BigInt(entry.votingPower ?? '0') > BigInt(0);
```
tyhonchik: worth using `bigIntUtils.safeParse`?
```suggestion
            const hasVotingPower = bigIntUtils.safeParse(entry.votingPower) > BigInt(0);

**Review by tyhonchik — COMMENTED** [review]
Hey! Nice work. I spotted a couple of issues though:

**Pagination gets confused** when you/delegate aren't on the first backend page. Prepending extra items into `DataListContainer` children bumps `childrenItemCount` above what `DataListRoot` expects — so the first "load more" does nothing, counters show "12 of 50", and page boundaries shift. The further you are by voting power rank, the longer it stays broken.

**Perf concerns:**
- `pinnedMemberParams` and `delegateMemberParams` are new objects every render — worth wrapping in `useMemo` so downstream hooks don't re-evaluate
- `hasValidDelegate` calls `addressUtils.isAddressEqual` on every render — also worth memoizing
- Three passes over the member list (`find` + `find` + `filter`) — may be replaced with a single pass using a `Set<string>` of pinned addresses for O(1) lookups
- All the fetching + computation logic is inline in the component body

**Suggested fix — prepend + trim:**
Build the merged list `[...pinned, ...rest]`, then `.slice(0, memberList.length)`. This keeps `childrenItemCount` exactly what DataList expects — trimmed tail items haven't been seen by the user anyway, and they show up naturally as more pages load. Single `Set`-based filter pass, O(n), scales the same for 50 or 10k members.

Also IMO makes sense to extract a `useTokenPinnedMembers` hook (all the connection/delegate/useMember logic, memoized internally) — the component doesn't pay for any of this computation, and the hook is independently testable. Plus use one `useMemo` in `TokenMemberList` for the merge + trim — the only work the component does, one pass, no extra allocations

File: src/plugins/tokenPlugin/components/tokenMemberList/tokenMemberList.tsx [hunk 4]
```diff
@@ -45,9 +135,16 @@ export const TokenMemberList: React.FC<ITokenMemberListProps> = (props) => {
                 layoutClassName="grid grid-cols-1 md:grid-cols-2 xl:grid-cols-3"
                 SkeletonElement={MemberDataListItem.Skeleton}
             >
-                {memberList?.map((member) => (
+                {mergedMemberList?.map((member) => (
```
milosh86: just a heads up, I also changed this component in https://github.com/aragon/app/pull/1027 ... also, there is now `LockToVoteMemberList` and `TokenMemberListBase` reused by both member list components.... I guess this logic will go into `TokenMemberListBase` then. Maybe to move the logic to a separate hook if it needs to be reused.

milosh86 (commented):
Just realised this is related to the other open PR, left a comment.

tyhonchik (dismissed):
looks great, thanks🤟

src/plugins/tokenPlugin/components/tokenMemberList/tokenMemberListBase.tsx
```diff
@@ -82,6 +89,53 @@ export const TokenMemberListBase: React.FC<ITokenMemberListBaseProps> = (
...
+        for (const member of memberList) {
+            const memberAddress = member.address.toLowerCase();
+            if (pinnedAddresses.has(memberAddress)) {
+                continue;
+            }
+
+            merged.push(member);
+        }
+
+        return merged.slice(0, memberList.length);
```
milosh86, line 136: why slice is needed here?

thekidnamedkd: the slice keeps the rendered child count equal to the backend page size.

thekidnamedkd: if memberList.length === 10 and neither the connected user nor delegate is in that page, the merged array becomes up to 12 items (2 pins, 10 from BE with the 2 already in there)

milosh86: ah, ok 👍

milosh86 (approved):
Great work 🚀

PR #1110, opened 2026-04-30
fix(APP-648): Strip hop-by-hop headers from proxied backend requests

milosh86:
# Description

Strips RFC 7230 hop-by-hop headers (`connection`, `keep-alive`, `proxy-authenticate`, `proxy-authorization`, `te`, `trailers`, `transfer-encoding`, `upgrade`, `host`) from requests before forwarding them to the upstream backend via the Next.js proxy route. Forwarding these headers can cause upstream servers to reject or mishandle requests.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/.changeset/fix-proxy-hop-by-hop-headers.md
+++ b/.changeset/fix-proxy-hop-by-hop-headers.md
@@ -1,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Strip hop-by-hop headers from proxied backend requests
--- a/src/modules/application/utils/proxyBackendUtils/proxyBackendUtils.test.ts
+++ b/src/modules/application/utils/proxyBackendUtils/proxyBackendUtils.test.ts
@@ -63,6 +63,25 @@ describe('proxyBackend utils', () => {
             expect(result.status).toEqual(204);
         });
 
+        it('does not forward hop-by-hop headers to upstream fetch', async () => {
+            const inboundHeaders = new Headers({
+                'transfer-encoding': 'chunked',
+                'keep-alive': 'timeout=5',
+                connection: 'keep-alive',
+                'content-type': 'text/plain',
+            });
+
+            await proxyBackendUtils.request(
+                generateNextRequest({ headers: inboundHeaders }),
+            );
+
+            const request = fetchSpy.mock.calls[0][1] as NextRequest;
+            expect(request.headers.get('transfer-encoding')).toBeNull();
+            expect(request.headers.get('keep-alive')).toBeNull();
+            expect(request.headers.get('connection')).toBeNull();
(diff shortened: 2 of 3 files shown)


thekidnamedkd:
Tracking this against [APP-648](https://linear.app/aragon/issue/APP-648). Fix shape looks right — the two empirically-undici-rejecting headers from a Node-side reproduction (`transfer-encoding`, `keep-alive`) are both in the strip list, so the Sentry signal should stop.

One ask + two minor observations:

**Ask: regression test for AC #4.** A small test in `proxyBackendUtils.test.ts` to lock the behavior so a future refactor of `buildRequestOptions` doesn't quietly bring it back:

```ts
it('does not forward hop-by-hop headers (transfer-encoding, keep-alive) to upstream fetch', async () => {
    const inboundHeaders = new Headers({
        'transfer-encoding': 'chunked',
        'keep-alive': 'timeout=5',
        connection: 'keep-alive',
    });

    await proxyBackendUtils.request(
        generateNextRequest({ url: 'http://test.com', headers: inboundHeaders }),
    );

    const outbound = fetchSpy.mock.calls[0][1] as RequestInit;
    const sent = new Headers(outbound.headers);
    expect(sent.get('transfer-encoding')).toBeNull();
    expect(sent.get('keep-alive')).toBeNull();
    expect(sent.get('connection')).toBeNull();
});
```

(Confirm `generateNextRequest` accepts a `headers` option; if not, build the `NextRequest` inline.)

**Minor observations (not blockers):**

- `content-length` isn't stripped. Body is re-read via `request.text()` and re-passed as a string, so Node should recompute it — but if an inbound `Content-Length` is preserved and somehow disagrees with the re-serialized body, undici trusts the explicit header. Belt-and-braces would add it to the strip list. Likely never observed in practice since inbound and outbound bodies are byte-identical here.
- `trailers` (plural) is in the list; the RFC 7230 name is `Trailer` (singular). HTTP/2 uses `trailers` in some contexts. Adding `trailer` too is one extra string for full coverage.

Empirical reproduction (plain Node, no Vercel) for reference:

```
[transfer-encoding: chunked] THROW TypeError: fetch failed
  cause: InvalidArgumentError: invalid transfer-encoding header
[keep-alive: timeout=5]      THROW TypeError: fetch failed
  cause: InvalidArgumentError: invalid keep-alive header
[connection: keep-alive]     OK status=200
[baseline]                   OK status=200
```

[APP-648]: https://aragonassociation.atlassian.net/browse/APP-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

thekidnamedkd (approved):
LGTM ✨

PR #1112, opened 2026-05-02
fix(APP-607): Allow gaugeVoter plugin to work with 3rd party escrow contracts

milosh86:
# Description

Fixes gauge voter pages failing when a DAO uses a 3rd party locker contract (e.g. Citrea) instead of the native veToken locker. In this setup, `ivotesAdapter` on the gauge voter contract returns the zero address, causing RPC-based voting power reads to silently fail.

**Changes:**
- `ivotesAdapterAddress` is now passed as a param from the call site (`plugin.settings.token.address`) rather than fetched via an additional RPC call inside `useGaugeVoterUserData`. This works for both the native and 3rd party locker cases.
- The Lock panel tab is hidden for 3rd party locker DAOs (where `plugin.meta.votingEscrow == null`), since locking is handled externally. Only the Delegate tab is shown.
- `fallbackValue` for the locks panel tab URL param is now derived from the first visible tab, so it correctly falls back to Delegate in the 3rd party locker case.

Closes [APP-607](https://linear.app/aragon/issue/APP-607)

## Type of Change

## Developer Checklist:

## Review Checklist:

[APP-607]: https://aragonassociation.atlassian.net/browse/APP-607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/plugins/tokenPlugin/components/tokenLockAndWrapOnboardingWatcher/tokenLockAndWrapOnboardingWatcher.tsx
```diff
@@ -72,11 +72,9 @@ export const TokenLockAndWrapOnboardingWatcher: React.FC<
         setHasPendingConnection(false);
 
         if (
-            // Either GaugeVoter or TokenVoting with a votingEscrow token adapter.
-            eligiblePlugin.interfaceType === PluginInterfaceType.GAUGE_VOTER ||
-            (eligiblePlugin.interfaceType ===
-                PluginInterfaceType.TOKEN_VOTING &&
-                (eligiblePlugin as ITokenPlugin).votingEscrow != null)
+            // Either GaugeVoter with native votingEscrow or TokenVoting with a votingEscrow token adapter.
```
thekidnamedkd, line 75: I think this still has a small edge case because we pick the first eligible plugin up front. If that first one is a GaugeVoter using a 3rd-party escrow, we now skip the lock onboarding, but we also never fall through to a later TOKEN_VOTING plugin that might still need the wrap onboarding. In a DAO with both plugin types, that means connect-onboarding could end up doing nothing here.

thekidnamedkd (commented):
One small comment. Nice job! 🌞

src/plugins/tokenPlugin/components/tokenLockAndWrapOnboardingWatcher/tokenLockAndWrapOnboardingWatcher.tsx
milosh86: yeah, true, but this is more related to onboarding logic, this PR was about making sure that app takes care properly the case when `votingEscrow` is not available in `gauge` plugin.

Regarding the onboarding logic, it was mentioned [in this ticket](https://linear.app/aragon/issue/APP-97/lock-and-wrap-onboarding-dialogs) as nice to have feature to support multiple plugins, but marked as highly unlikely, so it was left out as it would add complexity.

thekidnamedkd (approved):
niceeee

PR #1116, opened 2026-05-04
fix: Set the Condition factory address for Citrea

brickpop:
# Description

Setting the condition factory address of Citrea:
https://explorer.mainnet.citrea.xyz/address/0x150D61DdC7a6Df80fE2d3F7b7f58B36B69B25fdB?tab=contract


milosh86 (approved):
Added changeset, so we need additional approve now.

thekidnamedkd (approved):
🔥

PR #1117, opened 2026-05-04
feat(APP-649): Add support for existing proposal creation conditions in governance process

tyhonchik:
# Description

- Introduced new interface for existing proposal creation conditions.
- Updated form components to handle existing conditions, including UI for adding and removing conditions.
- Integrated feature flag for enabling existing condition functionality.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/src/assets/locales/en.json
+++ b/src/assets/locales/en.json
@@ -972,6 +972,12 @@
             "error": "You must select at least one body.",
             "label": "Bodies"
           },
+          "existingCondition": {
+            "add": "Existing condition",
+            "addressRequired": "Address is required",
+            "basicNotSupported": "Adding existing conditions is only available in Advanced governance.",
+            "remove": "Remove condition"
+          },
           "mode": {
             "anyDescription": "Any address can create proposals",
             "anyLabel": "Anyone",
--- a/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/existingConditionCard.tsx
+++ b/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/existingConditionCard.tsx
@@ -62,7 +62,7 @@ export const ExistingConditionCard: React.FC<IExistingConditionCardProps> = (
         onAddressChange(value?.address ?? '');
 
     return (
-        <Card className="w-full shrink-0 border border-neutral-100 p-4 shadow-neutral-sm">
+        <Card className="w-full shrink-0 border border-primary-400 p-4 shadow-primary">
             <div className="flex w-full min-w-0 flex-col items-start gap-3 sm:flex-row sm:items-start sm:gap-2">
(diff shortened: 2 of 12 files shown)


src/assets/locales/en.json
```diff
@@ -972,6 +972,12 @@
             "error": "You must select at least one body.",
             "label": "Bodies"
           },
+          "existingCondition": {
+            "add": "Existing condition",
+            "addressRequired": "Address is required",
+            "basicNotSupported": "Adding existing conditions is only available in Advanced governance.",
```
milosh86, line 978: Title mentions SPP but AC says it should be available on both basic and advanced processes. please double check with Evan if not already.

src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/existingConditionCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+     * Chain id used to build the explorer link for the address input.
+     */
+    chainId?: number;
+    /**
+     * Called when the user removes this card.
+     */
+    onRemove: () => void;
+}
+
+export const ExistingConditionCard: React.FC<IExistingConditionCardProps> = (
```
milosh86, line 31: I'm not 100% sure but there is also `<AddressesInputItem />` component which seems like it fits the need for this. @thekidnamedkd  created it recently to unify address lists. Please check, we might be able to reuse it.

milosh86 (commented):
Hey looks good, I left a couple of comments.

Also one more thing, a suggestion to consider. Maybe we split this into 2 separate sections: `Body` and `Existing conditions`? These are not really bodies, plus you cannot deselect body checkbox which seems unintuitive when you have other addresses in the list. WDYT? cc @evanaronson  also.

Image

src/assets/locales/en.json
evanaronson: Yes, but we can fix this later! This was a bit rushed to have it available for us in advanced processes.

src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/existingConditionCard.tsx
evanaronson: IMO we might want these separated out like Andrii has it here because it each one is a distinct condition. The mental model of collapsing them into one "object" seems wrong.

evanaronson:
I tested it and I think the biggest issue with the current visuals is it doesn't "included" with the rest.

IMO the following fixes would solve it:
- Put a blue border around it like the other selected conditions.
- Make the X button seconday.

@tyhonchik

milosh86 (approved):
looks good 👍

PR #1124, opened 2026-05-07
feat(APP-624): Implement guardrail skills for agents

thekidnamedkd:
# Description

MVP for path-scoped agent guardrails (APP-624). A lazy-loaded rule-skill system: narrow Markdown rules live alongside the codebase, and a shared loader injects only the rules whose `globs` field matches the file being edited — instead of carrying the whole project handbook in every prompt.

**Architecture (agent-neutral core + thin per-runtime adapter):**

- `.agents/hooks/inject-rules.mjs` — shared loader. Walks rule folders, matches globs, builds the rule stream. Two output adapters:
  - `generic` (default) — agent-neutral JSON: `{ relPath, matches, additionalContext }`. Any runtime that can shell out consumes this (Codex CLI, custom scripts, etc.).
  - `claude` — Claude Code's proprietary `hookSpecificOutput` JSON shape.
- `.claude/hooks/inject-rules.mjs` — small Claude adapter wrapping the shared loader for `PreToolUse`.
- `.agents/skills/rules/*.md` — checked-in shared rules.
- `.claude/skills/rules/*.md` — gitignored local IC layer (additive, never overrides shared).

**Rule-skills shipped (3 high-traffic areas, per AC):**

- `query-and-cache` — `src/**/api/**`, `src/**/queries/**`, `src/**/mutations/**`
- `plugin-slot-registration` — `src/plugins/*/index.ts`
- `rule-authoring` — `.agents/skills/rules/**`, `.claude/skills/rules/**` (dogfoods the system on its own folder)

**Wiring:**

- `.claude/settings.json` PreToolUse entry for `Edit|Write|MultiEdit`
- Root `CLAUDE.md` + `AGENTS.md` route to `.agents/skills/rules/`
- `.gitignore` selectively exposes only `.agents/hooks/`, `.agents/skills/rules/`, `.agents/metrics/`, `.claude/hooks/`, `.claude/settings.json`
- `pnpm test:guardrails` runs the loader + adapter contract tests via Node's built-in `node --test`

**Metrics pipeline (kill/keep telemetry for the spike):**

- `.agents/metrics/hits.jsonl` — shared, append-only event log. One row per (event × matched rule). Tracked in git.
- `.agents/metrics/.buffer.jsonl` — per-developer transient buffer, gitignored.
- `.husky/pre-commit` drains the buffer into the shared file on every commit; rows ride with each PR.
- `pnpm guardrails:stats` aggregates per-rule hits, p50/p95 latency, byte cost. All math in deterministic Node, never in an LLM.
- Full design + reliable-sum walkthrough: `.agents/metrics/README.md`.

**PR feedback addressed:**

- Renamed frontmatter field `applies-to` → `globs` (matches Cursor's convention; closer to a cross-harness norm)
- Fixed `canonicalize` test assertion (`endsWith('app-next')` → `existsSync(real)`) so it passes for any clone path
- Removed the blanket `ultracite fix` PostToolUse hook (was reformatting the whole repo after every edit)
- Fixed Claude adapter bug: `cli.adapter` defaulted to `'generic'`, so `cli.adapter || defaultAdapter` always picked generic — Claude Code was silently receiving wrong-shaped output and injecting nothing

**Inspect:**

```sh
# generic adapter (Codex / any runtime)
node .agents/hooks/inject-rules.mjs --file "$PWD/src/shared/api/daoService/queries/useDao.ts"

# Claude adapter (proprietary PreToolUse shape)
echo '{"tool_name":"Edit","tool_input":{"file_path":"'"$PWD"'/src/shared/api/daoService/queries/useDao.ts"}}' \
  | node .claude/hooks/inject-rules.mjs

pnpm test:guardrails
pnpm guardrails:stats
```

Spec: `.agents/skills/rules/README.md`. Adapter contract: `.agents/hooks/README.md`. Metrics: `.agents/metrics/README.md`.

## Type of Change

## Developer Checklist:

## Review Checklist:


tyhonchik (commented):
hey, really nice PR!

a few things:
- the `canonicalize` test has `assert.ok(real.endsWith('app-next'))`, but the repo is `aragon/app` — that assertion will fail for anyone who clones with the default folder name. maybe `existsSync(real)` instead?
- `PostToolUse: pnpm dlx ultracite fix` with no path runs against the whole repo after every single edit. could we pass the file from `tool_input.file_path` so it only touches what was just changed?
- `CLAUDE.md` and `AGENTS.md` are two nearly-identical files right now. the AGENTS.md spec convention is to symlink `CLAUDE.md` → `AGENTS.md` so they can't drift. intentional to keep them as two?


and one bigger thing to think about: the frontmatter (`applies-to`, `kind: rule`) is bespoke. Cursor Rules already has a native `globs:` field, and Claude Code Rules has `paths:` since 2.0.64. if we named the field one of those, the same `.md` files would be picked up natively without our loader. was that considered, or did we intentionally keep our own format (for future negation/alternation/etc)?

last thought - would be great to put some monitoring on this before we ship more rules: token count per session, latency added by the hook, agent response times before/after. without that we're flying blind on whether we're actually saving tokens (with 3 small rules + prompt cache the math isn't obviously a win yet) or whether we're slowing every Edit down. would hate to silently degrade DX or burn budget. could be as simple as logging matched-rule sizes + hook elapsed_ms to a file we can eyeball weekly.

.agents/shared/skills/rules/plugin-slot-registration.md
```diff
@@ -0,0 +1,62 @@
+---
+name: plugin-slot-registration
+description: Plugin index.ts wiring — registry chain, slot kinds, and the silent-failure modes that aren't visible from a single file.
+applies-to: src/plugins/*/index.ts
+kind: rule
```
milosh86, line 5: I think rules are on the same level as skills (not nested in skills): https://code.claude.com/docs/en/memory#organize-rules-with-claude/rules/

maybe nesting also works?

thekidnamedkd: Good question — kept them nested deliberately. There's no cross-harness "rules" location (Claude says `claude/rules/`, Cursor says `.cursor/rules/`, AGENTS.md has none), so picking one would lock us to that vendor's convention. `.agents/skills/rules/` piggybacks on skills as the de facto portable surface that most harnesses already scan. These aren't skills in the strict sense, but "rule-skill" was the least-bad framing for a guardrail that ships through the skills pipe. Open to revisiting once an actual cross-tool convention emerges.

CLAUDE.md
```diff
@@ -0,0 +1,49 @@
+# Aragon app-next
```
asciiman: Any reason why this needs to be a full copy instead of simply `@AGENTS.md`?

thekidnamedkd: Good catch! Thought this was already the case.

milosh86 (approved):
Looks good 👍

PR #1128, opened 2026-05-09
feat(APP-667): Pull token voting ERC20 members from aragon-subdomain

asciiman:
# Description

This repoints the mainnet ERC-20 Token Voting member list from the legacy backend to the aragon-domain package as the first slice of the platform architecture migration.

Requests are routed per plugin. Eligible token-voting queries go to the new domain BFF. Everything else falls back to the legacy backend unchanged. Both sources return the same library-owned DTO.

## What's in this PR

- **aragon-domain new serves member lists.** Mainnet plain-ERC-20 Token Voting membership is fetched from the aragon-domain through a new BFF route. Other networks, wrapped/VE tokens, lock-to-vote, multisig and admin keep using the legacy backend via a single routing predicate.
- **Package DTOs as the contract**: Member lists are typed with DTO pages from the aragon-domain regardless of source. Legacy responses are normalized through an anti-corruption mapper, and member activity now renders from indexed timestamps instead of per-block RPC lookups.
- **SSR hydration for member lists**: The members page prefetches the exact query key the token/lock-to-vote list consumes. There is a server options twin calling the domain controller in-process. The first paint comes from the server render instead of a client refetch.
- **Reusable domain-service plumbing**: The integration follows the `aragonDomainService` pattern previously introduced. A client/server service pair over a lazy server singleton now wires per-network RPC URLs and exposes the full controller via `getDomain()` instead of per-method delegates.

## Where to look

Start with `governanceService/utils/fetchTokenVotingMembership`. This is the single routing implementation both the client fetch and the RSC prefetch share, with the domain transport injected. `resolveMemberSource` holds the eligibility rules, and `buildTokenVotingMembershipParams` carries the server/client query-key contract. The two sides must build identical params.

## Notes

- `daoUtils.isLinkedAccountPlugin` now uses a server-safe address comparison. The gov-ui-kit shim's exports aren't callable from RSC context, and the members-page prefetch is its first server-side caller.
- Lock-to-vote uses the token-voting membership query because both lists render through `TokenMemberListBase`, and the routing predicate always sends LTV to the legacy backend. Moving LTV to its own query is deferred until the aragon-domain adds a lock-to-vote membership use case with its own DTO, so that code path only needs to be built once.

## Type of Change

## Developer Checklist:

## Review Checklist:


apps/app/package.json
```diff
@@ -30,7 +30,7 @@
         "e2e:codegen": "playwright codegen --config e2e/playwright.config.ts"
     },
     "dependencies": {
-        "@aragon/aragon-domain": "^0.3.1",
+        "@aragon/aragon-domain": "0.0.0-20260730113648",
```
asciiman: Don't merge until this is updated with release build of lib.

thekidnamedkd: Not approving yet, your own note still stands: the dep's on the dev snapshot `0.0.0-20260730113648`, so release build (and a rebase onto main) before this goes. Code side's solid, a couple of type nits inline.

apps/app/src/plugins/tokenPlugin/components/tokenProcessBodyField/tokenProcessBodyField.tsx
```diff
@@ -67,11 +67,21 @@ export const TokenProcessBodyField = (props: ITokenProcessBodyFieldProps) => {
...
     const initialParams = {
-        queryParams: { daoId, pluginAddress: isExisting ? body.address : '' },
+        queryParams: {
+            daoId,
+            pluginAddress: isExisting ? body.address : '',
+            network: dao?.network,
+            pluginInterfaceType: PluginInterfaceType.TOKEN_VOTING,
+            tokenAddress: membership.token.address,
+            tokenUnderlying:
+                (membership.token as { underlying?: string | null })
```
thekidnamedkd, line 77: This cast is quietly doing the routing. `ITokenSetupMembershipForm.token` doesn't declare `underlying`, and the value only survives because `tokenBodyUtils` spreads the full settings token in. If someone later builds `membership.token` to the declared shape, a mainnet wrapped or VE body flips to the domain source and shows the wrong holder count with no warning. Can we declare `underlying` on the form token type (picked off `ITokenPluginSettingsToken`) so the compiler catches it?

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
```diff
@@ -0,0 +1,101 @@
...
+    const interfaceType = queryParams.pluginInterfaceType;
+    const tokenUnderlying = queryParams.tokenUnderlying;
+
+    const useDomain =
+        pluginAddress != null &&
+        tokenAddress != null &&
+        network != null &&
+        DOMAIN_NETWORKS.has(network) &&
+        interfaceType === PluginInterfaceType.TOKEN_VOTING &&
+        tokenUnderlying == null;
```
thekidnamedkd, line 52: The VE exclusion leans on VE tokens always carrying `underlying`, but the type only promises that for the wrap-needed case, and `votingEscrow` never reaches the routing params. If a mainnet VE deployment ever ships with `underlying` null, its list reads plain `ERC20Votes` balances from the domain instead of escrow locks. Threading a `votingEscrow` flag through `buildTokenVotingMembershipParams` and excluding on it here keeps the predicate self-contained.

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+export const fetchTokenVotingMembership = async (
+    params: IGetTokenVotingMembershipParams,
+    fetchDomainMembers: FetchDomainTokenVotingMembers,
+    fetchLegacyMemberList: FetchLegacyMemberList,
+): Promise<PageDTO<TokenVotingMemberDTO>> => {
+    const { queryParams } = params;
+    const { tokenAddress, page, pageSize } = queryParams;
+
+    if (tokenAddress != null && resolveMemberSource(queryParams) === 'domain') {
```
thekidnamedkd, line 73: When the domain fetch throws, mainnet lists now hard-error where legacy used to serve them, since this slice has no runtime fallback. If that's on purpose so indexer problems surface instead of hiding behind legacy data, all good.

apps/app/src/modules/governance/pages/daoMembersPage/index.ts
```diff
@@ -1,2 +1,3 @@
-export { DaoMembersPage, type IDaoMembersPageProps } from './daoMembersPage';
+// The DaoMembersPage RSC is NOT exported here on purpose since imports
```
thekidnamedkd, line 1: nit: missing word in the comment.
```suggestion
// The DaoMembersPage RSC is NOT exported here on purpose since it imports
```

milosh86 (commented):
@asciiman this one looks good overall ... could you please sync it with the latest main and address open comments, so that we could take a final pass and merge.

apps/app/package.json
tyhonchik: one more thing to fold into the release: the snapshot's query still reads `MemberGovernanceMetrics`, which was renamed to `MemberMetrics` in the indexer schema — so the domain branch 500s against the live indexer right now (left a comment on aragon/aragon-domain#4). release build should come after that lands.

apps/app/src/app/api/domain/token-voting/members/route.ts
```diff
@@ -0,0 +1,82 @@
+import { type NextRequest, NextResponse } from 'next/server';
+// biome-ignore lint/style/noRestrictedImports: server-only BFF validation; strict EIP-55 validation is explicitly disabled.
+import { isAddress } from 'viem';
+import { tokenVotingMembershipServiceServer } from '@/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipService.server';
+import { monitoringUtils } from '@/shared/utils/monitoringUtils';
+
+const maximumPageSize = 250;
+
+interface ITokenVotingMembershipQueryParams {
```
tyhonchik: no network param anywhere in this chain (route → server service → domain), and the domain query isn't chain-scoped either (raised on aragon/aragon-domain#4) — the same token address on two chains would mix members. once the domain takes a chainId, this request shape needs it too; maybe worth adding the param now to avoid a second API change.

apps/app/src/modules/governance/hooks/useTokenVotingMembershipData/useTokenVotingMembershipData.ts
```diff
@@ -0,0 +1,65 @@
...
+import type { InfiniteQueryOptions } from '@/shared/types';
+import { dataListUtils } from '@/shared/utils/dataListUtils';
+
+/**
+ * Gets the membership for token-voting plugins.
+ *
+ * This routes between the aragon-domain BFF and the legacy backend
+ * regardless of source. The generic `getMemberList` still serves multisig/admin.
+ */
+export const useTokenVotingMembershipData = (
```
milosh86: I just realised that we should probably move this and other token specific logic to tokenVoting plugin folder. It breaks plugin encapsulation a bit, as modules should not have knowledge about specifics of plugins. ... but let's not block this PR with this, we can do this as a separate task.

apps/app/package.json
tyhonchik: still on the old snapshot for the moment — the new one (aragon/aragon-domain#4 with chainId in the request DTO) is blocked on a CI secret in the domain repo, so type-check here is red until the bump. switching to `^0.4.0` once #4 is released; merge waits on that.

apps/app/src/plugins/tokenPlugin/components/tokenProcessBodyField/tokenProcessBodyField.tsx
tyhonchik: declared `underlying` on the form token type (picked off `ITokenPluginSettingsToken`), cast is gone. optional rather than required: the create/import wizard writes the token field by field and never has an underlying; existing bodies get it from the `tokenBodyUtils` spread. routing doesn't lean on it alone anymore, see the escrow thread.

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
tyhonchik: threaded an explicit `hasVotingEscrow` through the params (from `plugin.settings.votingEscrow`; the body field reads the installed plugin's settings since the form doesn't carry it) and the predicate excludes on it independently of `underlying`.

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
tyhonchik: changed: the domain branch now falls back to the legacy backend when the call fails (error logged, page served from the backend). the ui shouldn't depend on the domain being up — it's just a switch at the ssr level. the switch itself is the new `domainMemberList` flag, off everywhere for now.

apps/app/src/modules/governance/pages/daoMembersPage/index.ts
tyhonchik: fixed.

apps/app/src/app/api/domain/token-voting/members/route.ts
tyhonchik: route takes `chainId` now (required, positive int), the routing layer maps the dao network to it, and the domain request DTO carries it (aragon/aragon-domain#4).

apps/app/src/modules/governance/hooks/useTokenVotingMembershipData/useTokenVotingMembershipData.ts
tyhonchik: agreed — separate ticket: move `useTokenVotingMembershipData`, `buildTokenVotingMembershipParams`, `fetchTokenVotingMembership` and the routing predicate under `plugins/tokenPlugin`, module reaches them through the plugin definition.

tyhonchik:
synced with main and pushed the review fixes in one commit (2bdde088): `domainMemberList` flag (off everywhere), legacy fallback on domain errors, chainId on the bff route, explicit escrow exclusion. answers inline. type-check stays red until the aragon-domain snapshot from #4 lands here — bump follows.

apps/app/package.json
tyhonchik: bumped to snapshot `0.0.0-20260908085601` (`snapshot-34207155965`, built from aragon/aragon-domain#4). `^0.4.0` once #4 is released.

apps/app/src/app/api/domain/token-voting/members/route.ts
```diff
@@ -0,0 +1,83 @@
...
+import { isAddress } from 'viem';
+import { tokenVotingMembershipServiceServer } from '@/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipService.server';
+import { monitoringUtils } from '@/shared/utils/monitoringUtils';
+
+const maximumPageSize = 250;
+
+const isPositiveInteger = (value: number): boolean =>
+    Number.isInteger(value) && value > 0;
+
+const validateQueryParams = (
```
thekidnamedkd: nit: projectStructure.md:10 keeps logic out of /app, and both sibling routes hold to it (api/rpc is a one-line delegate, member-profile is a bare passthrough). Moving validateQueryParams and the page cap next to tokenVotingMembershipService would also give the supported-chain check somewhere to live.

apps/app/src/app/api/domain/token-voting/members/route.ts
```diff
@@ -0,0 +1,83 @@
...
+    const tokenContractAddress = params.get('tokenContractAddress');
+    const page = params.get('page');
+    const pageSize = params.get('pageSize');
+    const parsedChainId = chainId != null ? Number(chainId) : undefined;
+    const parsedPage = page != null ? Number(page) : undefined;
+    const parsedPageSize = pageSize != null ? Number(pageSize) : undefined;
+
+    if (
+        parsedChainId == null ||
+        !isPositiveInteger(parsedChainId) ||
```
thekidnamedkd: chainId only has to be a positive integer here, but the controller is built with RPC urls for mainnet only (aragonDomainService.backend.ts:13), so ?chainId=137 goes to Envio with nothing behind it and comes back as a logged 500. Could we validate against the supported chain set like proxyRpcUtils.chainIdToRpcEndpoint does?

apps/app/src/app/api/domain/token-voting/members/route.ts
```diff
@@ -0,0 +1,83 @@
...
+    return {
+        chainId: parsedChainId,
+        pluginAddress,
+        tokenContractAddress,
+        page: parsedPage,
+        pageSize: parsedPageSize,
+    };
+};
+
+export const GET = async (req: NextRequest) => {
```
thekidnamedkd: The flag gates the callers but not this route, so from deploy anyone can drive Envio and our RPC through it while the app itself never calls it. Since the whole point is that merging changes nothing yet, should the handler check domainMemberList too and 404 when it is off?

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export type FetchLegacyMemberList = (
+    params: IGetMemberListParams,
+) => Promise<IPaginatedResponse<ITokenMember>>;
+
+/**
+ * Networks whose token-voting member queries are served by the
+ * aragon-domain BFF. Expand as more networks are indexed by aragon-indexer.
+ */
+const DOMAIN_NETWORKS: ReadonlySet<Network> = new Set([
```
thekidnamedkd: DOMAIN_NETWORKS here and domainRpcNetworks in aragonDomainService.backend.ts:13 are the same fact in two places. Add Base to this one and forget the other and every Base call fails into the legacy fallback with only a Sentry line to show it. Could export one constant and have both read it.

apps/app/src/plugins/tokenPlugin/components/tokenProcessBodyField/tokenProcessBodyField.tsx
```diff
@@ -61,16 +63,39 @@ export const TokenProcessBodyField = (props: ITokenProcessBodyFieldProps) => {
...
+        daoId,
+        pluginAddress: isExisting ? body.address : '',
+        includeSubPlugins: true,
+        includeLinkedAccounts: true,
+    })?.[0];
+    const existingPluginSettings = existingPlugin?.meta.settings as
+        | ITokenPluginSettings
+        | undefined;
+
     const initialParams = {
```
thekidnamedkd: This rebuilds the routing params by hand and casts the settings to get at votingEscrow, while buildTokenVotingMembershipParams plus isTokenMemberListPlugin do both and are the documented single source for the param shape. Next routing input we add gets threaded through the helper and quietly misses this site.

apps/app/src/modules/governance/api/governanceService/governanceService.ts
```diff
@@ -52,6 +57,21 @@ class GovernanceService extends AragonBackendService {
         return result;
     };
 
+    /**
+     * Gets the membership for token-voting plugins.
+     *
+     * This routes between the aragon-domain BFF and the legacy backend
+     * regardless of source. The generic `getMemberList` still serves multisig/admin.
+     */
+    getTokenVotingMembership = (
```
thekidnamedkd: AC1 on APP-667 asks for the backend/domain decision to happen in the BFF, but resolveMemberSource runs client-side here and again in the server options, with the route as a pure passthrough. Fine by me as a design, just worth settling explicitly so the ticket is not left claiming a seam we did not build.

apps/app/src/modules/governance/hooks/useTokenVotingMembershipData/useTokenVotingMembershipData.ts
```diff
@@ -0,0 +1,65 @@
...
+    params: IGetTokenVotingMembershipParams,
+    options?: InfiniteQueryOptions<
+        PageDTO<TokenVotingMemberDTO>,
+        IGetTokenVotingMembershipParams
+    >,
+) => {
+    const { t } = useTranslations();
+
+    const {
```
thekidnamedkd: nit: this is useMemberListData copied line for line except the query, error and empty state strings included. Could one take the query options and the other go away, or is that better folded into the tokenPlugin move you and Milos agreed on?

apps/app/src/modules/governance/api/governanceService/queries/useTokenVotingMembership/useTokenVotingMembership.ts
```diff
@@ -0,0 +1,35 @@
...
+> => ({
+    queryKey: governanceServiceKeys.tokenVotingMembership(params),
+    initialPageParam: params,
+    queryFn: ({ pageParam }) =>
+        governanceService.getTokenVotingMembership(pageParam),
+    getNextPageParam: governanceService.getNextPageParams,
+    ...options,
+});
+
+export const useTokenVotingMembership = (
```
thekidnamedkd: nit: nothing imports useTokenVotingMembership and the barrel only re-exports the options, so this hook is dead on arrival. Same story for export type AragonDomainController.

thekidnamedkd (commented):
Posture: COMMENT, nothing blocking from my pass. The domainMemberList gate is off everywhere so the merge is behavior-neutral; the real merge gate is the aragon-domain snapshot vs ^0.4.0 thread, and that looks like a dependency release chore more than a code change. CI is green at head including type-check. Comments below, two on the new BFF route (unsupported chain ids, route not behind the flag), the duplicated network constant, a third site building routing params by hand, and nits. Also replied on the fallback thread about falling back whole-query instead of per page.

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
thekidnamedkd: One thing the fallback opens up: it is per page, so a blip on page 3 leaves pages 1-2 from the domain and page 3 onward from the backend, with getNextPageParams continuing off whichever metadata came back last. Different ordering and counts between sources means duplicated or skipped members mid-scroll. Falling back for the whole query (or only on the first page) would keep one list from one source.

apps/app/src/app/api/domain/token-voting/members/route.ts
tyhonchik: moved: validation and the page cap sit next to the service, the route is a two-liner like api/rpc

apps/app/src/app/api/domain/token-voting/members/route.ts
tyhonchik: moot now: the client doesn't send a chain at all. the bff takes the network from the dao and only builds a domain request for networks the domain serves — anything else goes to the backend.

apps/app/src/app/api/domain/token-voting/members/route.ts
tyhonchik: the route is the members endpoint now, not a domain passthrough. it reads domainMemberList server-side and that alone decides domain vs backend, so nothing reaches envio unless the flag is on for that request.

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
tyhonchik: one domainNetworks constant in aragonDomainService, read by the controller's rpc urls and by the routing chec

apps/app/src/plugins/tokenPlugin/components/tokenProcessBodyField/tokenProcessBodyField.tsx
tyhonchik: gone now
the component just asks for the members of a plugin (daoId + pluginAddress). no settings cast, no routing fields to keep in sync.

apps/app/src/modules/governance/api/governanceService/governanceService.ts
tyhonchik: fair point, fixed
the decision now lives in tokenVotingMembershipService.server: it resolves the dao and plugin, reads the flag and picks the source. the client and the rsc prefetch both just ask for a plugin's members — the query key is daoId/pluginAddress/paging, and nothing in the ui can tell (or influence) which source answered. the rsc path calls the same service in-process. AC1 holds as written.

apps/app/src/modules/governance/hooks/useTokenVotingMembershipData/useTokenVotingMembershipData.ts
tyhonchik: still folding it into the plugins/tokenPlugin move Milos and i agreed on

apps/app/src/modules/governance/api/governanceService/queries/useTokenVotingMembership/useTokenVotingMembership.ts
tyhonchik: dropped useTokenVotingMembership
AragonDomainController is internal to the backend service now

apps/app/src/modules/governance/api/governanceService/utils/fetchTokenVotingMembership/fetchTokenVotingMembership.ts
tyhonchik: per page is gone
the bff decides once: only a first page may fall back, a mid-list domain failure errors instead of splicing backend data into a domain list, and a list that started on the backend stays there — the page carries its source and the next request echoes it back as an opaque continuation.

tyhonchik:
@milosh86 @thekidnamedkd reworked the routing after this round: the backend/domain decision happens in the bff now, not in the client.

- the ui asks for a plugin's members and gets a page — no network, token, escrow or flag in the query, and no way to tell which source answered.
- `tokenVotingMembershipService.server` resolves the dao and plugin, reads `domainMemberList` server-side and picks the source;
- the members-page prefetch calls that same service in-process;
- the route (/api/governance/token-voting/members) is a delegate with its validation next to the service.
- fallback is a whole-list decision, the network list is one constant, dead exports are gone. 

merge still waits on aragon-domain#4 being released so we can bump off the snapshot

apps/app/src/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipService.server.ts
```diff
@@ -0,0 +1,141 @@
...
+class TokenVotingMembershipServiceServer {
+    getTokenVotingMembership = async ({
+        queryParams,
+    }: IGetTokenVotingMembershipParams): Promise<ITokenVotingMembershipPage> => {
+        const { daoId, pluginAddress, page, pageSize, source } = queryParams;
+        const isFirstPage = page == null || page === 1;
+
+        // `source` pins a list that already started on the legacy backend, so
+        // a recovering domain cannot swap sources mid-scroll.
+        const domainRequest =
```
thekidnamedkd: I like this a lot, the BFF owning the source and the page carrying it back is much nicer than where this started. One thing I think the pinning misses: it holds a backend list, but `source: 'domain'` on page 3 re-runs `resolveDomainRequest`, and that can come back undefined without throwing, either because the flag flipped off between pages or the `getDao` read failed into the catch at 128. Then we fall through to the backend branch and a backend page lands in a domain list. The `!isFirstPage` throw would catch it but it's inside the `domainRequest != null` branch so it never sees this path. What if a `source: 'domain'` request past page 1 threw when the domain request won't resolve? Your tests actually sit either side of it, 175 has the domain call failing mid-list and 186 has the DAO failing on page 1.

apps/app/src/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipRouteUtils.ts
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        try {
+            const result =
+                await tokenVotingMembershipServiceServer.getTokenVotingMembership(
+                    { queryParams },
+                );
+
+            return NextResponse.json(result);
+        } catch (error) {
```
thekidnamedkd: Small thing I noticed: when the service rethrows a mid-list failure it's already logged it as `token_voting_membership_domain_error`, so we get a second event here under a different name with the same daoId/pluginAddress/page. Might make the Sentry counts look worse than they are. Maybe the service only logs where it swallows and this handler covers whatever comes back up?

apps/app/src/modules/governance/api/governanceService/governanceServiceKeys.ts
```diff
@@ -41,6 +43,10 @@ export const governanceServiceKeys = {
         GovernanceServiceKey.MEMBER_LIST,
         params,
     ],
+    tokenVotingMembership: (params: IGetTokenVotingMembershipParams) => [
```
thekidnamedkd: Since the service got its own folder, would it make sense for the key and the `queries/useTokenVotingMembership` options to move over with it? Right now `governanceServiceKeys.ts` type-imports from `tokenVotingMembershipService` while the service imports `governanceService`, and having them together would untangle that. `dataFetching.md` keeps keys and `/queries` in the service folder too, so it'd line up. Happy either way if you'd rather the query stay owned by governanceService.

apps/app/src/modules/governance/api/governanceService/queries/useTokenVotingMembership/useTokenVotingMembership.ts
```diff
@@ -0,0 +1,28 @@
...
+    ITokenVotingMembershipPage,
+} from '@/modules/governance/api/tokenVotingMembershipService';
+import { tokenVotingMembershipServiceClient } from '@/modules/governance/api/tokenVotingMembershipService';
+import type {
+    InfiniteQueryOptions,
+    SharedInfiniteQueryOptions,
+} from '@/shared/types';
+import { governanceServiceKeys } from '../../governanceServiceKeys';
+
+export const tokenVotingMembershipOptions = (
```
thekidnamedkd: nit: now that the hook's gone this folder's the only one under `queries/` not exporting the `use*` it's named for, just the options pair. `tokenVotingMembershipOptions` would read truer, though it sorts itself out if these move next to the service anyway.

apps/app/src/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipService.server.ts
tyhonchik: `source: 'domain'` past page 1 now throws when the domain request won't resolve (flag off, dao read failed, plugin out of coverage) instead of falling through to the backend. split the service into `getDomainMembership` (undefined = not covered, throw = failed) and `getBackendMembership`, the pin check sits between them. tests for both mid-list cases added.

apps/app/src/modules/governance/api/tokenVotingMembershipService/tokenVotingMembershipRouteUtils.ts
tyhonchik: done. the service logs only where it swallows (first-page fallback), whatever propagates is logged once by the route. tests assert a single log on each path.

apps/app/src/modules/governance/api/governanceService/governanceServiceKeys.ts
tyhonchik: moved. `tokenVotingMembershipServiceKeys` and `queries/useTokenVotingMembership` (options, hook, server twin) live in the service folder, `governanceServiceKeys` no longer imports from it.

apps/app/src/modules/governance/api/tokenVotingMembershipService/queries/useTokenVotingMembership/useTokenVotingMembership.ts
```diff
@@ -0,0 +1,28 @@
...
+    ITokenVotingMembershipPage,
+} from '@/modules/governance/api/tokenVotingMembershipService';
+import { tokenVotingMembershipServiceClient } from '@/modules/governance/api/tokenVotingMembershipService';
+import type {
+    InfiniteQueryOptions,
+    SharedInfiniteQueryOptions,
+} from '@/shared/types';
+import { governanceServiceKeys } from '../../governanceServiceKeys';
+
+export const tokenVotingMembershipOptions = (
```
tyhonchik: sorted itself out with the move: the folder exports `useTokenVotingMembership` again and `useTokenVotingMembershipData` goes through it, same shape as `useMemberList` / `useMemberListData`.

thekidnamedkd (approved):
LGTM 🛰️

PR #1129, opened 2026-05-11
fix(APP-665): Fix wallet connection by syncing state with AppKit and initializing AppKit early in providers

milosh86:
## Description

Migrates wallet connect/disconnect flows to use AppKit utilities and updates the `useConnected` hook to rely on AppKit state. Also removes session persistence from the address impersonation feature.

Changes:
- Init AppKit in providers and use AppKit `connect`/`disconnect` utils in dialogs and navigation components
- Add `useWalletConnected` hook and `useConnectedWalletGuard` that rely on AppKit state
- Remove `localStorage` persistence from `useAddressImpersonation` (impersonator flag is now session-only)
- Fix slow WC QR code display
- Remove Reown analytics and other unused features

Closes APP-665

## Type of Change

## Developer Checklist:

## Review Checklist:


milosh86 (commented):
Self-review completed, ready for team review...

tyhonchik (commented):
Hey, a few small ideas not blockers

1. CSP in proxyUtils.ts
`allowedInFrameHosts` is "who can embed our app in an iframe" — i.e. frame-ancestors. But the variable is reused for both `frame-src` and `frame-ancestors`. So adding `https://verify.walletconnect.org` there also lets WC embed our app which we don't actually need; we only need to embed their verify iframe.

Not dangerous, but the name now lies. Maybe split into `iframeEmbedders` (-> frame-ancestors) and `embeddableHosts` (-> frame-src)? WC verify only needs the second one.

2. Short window where address and `isConnected` can mismatch in nav components

In `NavigationDao`, `NavigationWizard`, `ExploreNav`, `DashboardOnboarding` we now do:

```ts
const { address } = useConnection();        // wagmi
const isConnected = useWalletConnected();   // AppKit
```

`DesyncWatcher` waits 1.5s before forcing a disconnect, and inside that window AppKit can say connected while wagmi hasn't surfaced an address yet (or vice versa). UI renders the "connected" pill but the address inside it is empty.

Maybe it worth doing: `effectiveIsConnected = isMounted && isConnected && address != null`

3. Idea - ban `isConnected` from `useConnection` via Biome
So nobody accidentally bypasses the new `useWalletConnected`. Biome can't ban one destructured field, but we can do what we already do for `useEnsName`/`useEnsAvatar`:

- Add a thin wrapper, e.g. `useWalletAccount`, returning only what we actually want from wagmi (`address`, `chainId`, `isReconnecting`) — no `isConnected`.
- Migrate existing call sites to `useWalletAccount`
- Add `useConnection` to `noRestrictedImports` in `biome.json`
- Leave `DesyncWatcher` as the only legitimate useConnection consumer

.changeset/tasty-maps-stop.md
```diff
@@ -0,0 +1,5 @@
+---
```
thekidnamedkd, line 1: Why two changesets?

turbo.json
```diff
@@ -1,5 +1,6 @@
 {
     "$schema": "https://turbo.build/schema.json",
+    "concurrency": "70%",
```
thekidnamedkd: This is for what?

src/modules/application/components/desyncWatcher/desyncWatcher.tsx
```diff
@@ -0,0 +1,49 @@
...
+        ) {
+            return;
+        }
+        if (appKitConnected === wagmiConnected) {
+            return;
+        }
+
+        const timeoutId = setTimeout(() => {
+            void disconnect();
+        }, settleWindowMs);
```
thekidnamedkd, line 43: This looks like it will disconnect valid wagmi-only sessions, not just true AppKit/wagmi desyncs. One existing example is useAddressImpersonation, which connects through wagmi’s mock(...) connector directly, so wagmiConnected becomes true while AppKit still reports disconnected. In that case this effect will treat the session as a permanent mismatch and call disconnect() after 1.5s. Could we exempt non-AppKit connectors here, or otherwise distinguish “AppKit not involved” from a real desync?

src/modules/application/hooks/useWalletConnected/useWalletConnected.ts
```diff
@@ -0,0 +1,16 @@
...
+ * renders against, so the two cannot drift apart.
+ *
+ * Address and chainId reads should stay on wagmi (`useWalletAccount`),
+ * since web3 operations need the wagmi connector identity.
+ */
+export const useWalletConnected = (): boolean => {
+    const { isConnected } = useAppKitAccount({ namespace: 'eip155' });
+
+    return isConnected;
+};
```
thekidnamedkd, line 16: Using AppKit as the only source of truth for isConnected seems to hide supported wagmi-only sessions. We already have one in useAddressImpersonation, where wagmi has an address/connector but AppKit never created the session, so this hook still returns false. That means downstream guards and wallet CTAs will treat the user as disconnected even though wagmi is connected. Could this fall back to wagmi for non-AppKit sessions instead of relying only on useAppKitAccount()?

thekidnamedkd (commented):
Nice one. A few questions. 🦋

.changeset/tasty-maps-stop.md
milosh86: kinda 2 different issues and the WC QR code one was really small to extract to a separate PR.

turbo.json
milosh86: when tests are run with 100% of workers/threads it's actually running much slower ... this should've been Turbo's alternative for Jest's ` maxWorkers: '70%'`.

Anyways, this doesn't work as expected, I still get really slow test runs before git push, so I can remove this.

src/modules/application/components/desyncWatcher/desyncWatcher.tsx
milosh86: The point was to make sure those are always in sync, there should be no valid wagmi session without AppKit, right? AppKit should be the primary interface.

impersonator is a kind of hack using wagmi just because AppKit does not offer an API to something similar. I tested impersonator and it works fine. Did you managed to get it disconnected during a session?

btw, I had to remove persistance for impersonator. It was not just because of this logic, but I had some other issues too, I can't remember now. It was not worth adding complexity.

src/modules/application/hooks/useWalletConnected/useWalletConnected.ts
milosh86: As I mentioned above, I don't think there should be wagmi-only sessions. Even if you create a session through wagmi, AppKit will still pick it up and create it's own session, but it will be very unstable way of using it. That's totally fine for impersonation, but I don't see other use cases which would need to go around AppKit.

.changeset/focused-dialogs-help.md
tyhonchik, line 1: this shouldn't be deleted?

milosh86: true, it's from another PR, I don't know why it's deleted here .... good catch.

milosh86: done

tyhonchik (approved):
lgtm!

PR #1134, opened 2026-05-15
feat(APP-628): Fetch real-time token total supply for LTV participation and settings

milosh86:
## Description

Fixes the Lock-to-Vote participation denominator and settings display by fetching the ERC-20 `totalSupply` in real time instead of relying on stale values stored in the API response.

Changes:
- Extract `resolveServerRpcUrl` from `proxyRpcUtils` into its own module for reuse
- Add `fetchTokensTotalSupply` utility — fetches `totalSupply` for multiple token addresses; uses private RPC on the server and wagmi multicall on the client
- Add `collectTokenAddresses` utility to collect relevant token addresses from a proposal
- Update `governanceService.getProposalList` and `getProposalBySlug` to attach live `tokensTotalSupply` to each proposal
- Add `useTokenTotalSupply` shared hook for single-token supply queries
- Update `useLockToVoteGovernanceSettings` to fetch real-time supply via `useTokenTotalSupply`
- Update `lockToVoteProposalVotingBreakdown` and `lockToVoteSettingsUtils.parseSettings` to use real-time supply
- Add `tokensTotalSupply` support to SPP proposals

Closes APP-628

## Type of Change

## Developer Checklist:

## Review Checklist:


src/plugins/lockToVotePlugin/components/lockToVoteProposalVotingSummary/lockToVoteProposalVotingSummary.tsx
tyhonchik, line 81: After this PR `getProposalTokenTotalSupply` returns strictly `tokensTotalSupply[address]` with no fallback. If the address isn't present in the map (token-address case mismatch, stale React Query cache from before the schema change, partial RPC decoration), this becomes undefined - safeParse - 0n - '0' - 0, and the invariant throws — crashing the summary of an active proposal.

Previously `historicalTotalSupply` was always on the wire for active proposals, so the invariant never fired in practice. Could we either soft-fallback (render — / hide the section) or fail upstream in `attachTotalSupply` with a clear message so the page falls into Page.Error instead of a half-rendered crash?

src/modules/governance/api/governanceService/governanceService.ts
```diff
@@ -74,18 +78,62 @@ class GovernanceService extends AragonBackendService {
...
 
-        return result;
+        const addresses = Array.from(
+            new Set(result.data.flatMap(collectTokenAddresses)),
+        );
+        const tokensTotalSupply =
+            addresses.length === 0 || result.data.length === 0
+                ? {}
+                : await fetchTokensTotalSupply(
+                      result.data[0].network,
```
tyhonchik, line 95: `result.data[0].network` is used for all addresses on the page. Works today because the list is scoped by `pluginAddress`, but it's an invisible invariant — a future multi-DAO/multi-network feed will silently return wrong supplies. Either add a comment, or group addresses by network and run one multicall per network.

src/modules/application/utils/proxyRpcUtils/resolveServerRpcUrl.ts
```diff
@@ -0,0 +1,97 @@
...
+    [RpcProvider.ALCHEMY]: 'NEXT_SECRET_RPC_KEY',
+    [RpcProvider.ANKR]: 'NEXT_SECRET_ANKR_RPC_KEY',
+    [RpcProvider.DRPC]: 'NEXT_SECRET_DRPC_RPC_KEY',
+    [RpcProvider.PEAQ]: 'NEXT_SECRET_PEAQ_QUICKNODE_RPC_KEY',
+};
+
+/**
+ * Resolves the upstream RPC URL for a given network. The returned URL embeds the provider API
+ * key (`process.env.NEXT_SECRET_*`) as a path segment — it is a secret and must never be sent
+ * to the client. Callers MUST gate invocation behind `process.env.NEXT_RUNTIME` (build-time-
```
tyhonchik, line 18: doc says "callers MUST gate via `process.env.NEXT_RUNTIME` for tree-shake", but fetchTokensTotalSupply uses `typeof window` — a runtime check, so the bundler keeps the module in the client chunk. No secrets leak, but no tree-shake either. 

suggesting update the doc or switch the gate to `process.env.NEXT_RUNTIME === 'nodejs'` + dynamic `import()`.

src/plugins/lockToVotePlugin/components/lockToVoteProcessBodyField/lockToVoteProcessBodyField.tsx
```diff
@@ -106,6 +114,20 @@ export const LockToVoteProcessBodyField = (
...
+    const parsedMinParticipationToken = formatUnits(
+        bigIntUtils.safeParse(minParticipationToken),
+        tokenDecimals,
+    );
+    const formattedMinParticipationToken = formatterUtils.formatNumber(
+        parsedMinParticipationToken,
+        {
+            format: NumberFormat.TOKEN_AMOUNT_SHORT,
+        },
+    );
```
tyhonchik: Same formula in lockToVoteSettingsUtils.tsx. Worth extracting a shared helper?

src/plugins/lockToVotePlugin/components/lockToVoteProcessBodyField/lockToVoteProcessBodyField.tsx
```diff
@@ -86,14 +87,21 @@ export const LockToVoteProcessBodyField = (
         name: tokenName,
         symbol: tokenSymbol,
         decimals: tokenDecimals,
-        totalSupply,
     } = membership.token;
+
+    const { data: totalSupply } = useTokenTotalSupply({
+        chainId: chainId as number,
+        address: tokenAddress as Hex,
```
tyhonchik: Both fields are really `T | undefined`. Safe at runtime thanks to `enabled`, but the cast lies to the compiler. 

Suggestion: make `address?` / `chainId?` optional in `IUseTokenTotalSupplyParams` and auto-disable inside the hook when they're missing — kills the casts and the manual `enabled` plumbing.

src/plugins/lockToVotePlugin/hooks/useLockToVoteGovernanceSettings/useLockToVoteGovernanceSettings.ts
```diff
@@ -13,6 +16,17 @@ export const useLockToVoteGovernanceSettings = (
     const { settings, isVeto } = params;
 
     const { t } = useTranslations();
+    const { chainId } = useDaoChain({ network: settings.token.network });
+    const { data } = useTokenTotalSupply({
+        chainId: chainId as number,
+        address: settings.token.address as Hex,
```
tyhonchik: the same 
both fields are really `T | undefined`. Safe at runtime thanks to `enabled`, but the cast lies to the compiler. 

Suggestion: make `address?` / `chainId?` optional in `IUseTokenTotalSupplyParams` and auto-disable inside the hook when they're missing — kills the casts and the manual `enabled` plumbing.

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
```diff
@@ -38,20 +43,17 @@ class LockToVoteSettingsUtils {
     parseSettings = (
         params: IParseLockToVoteSettingsParams,
     ): IDefinitionSetting[] => {
-        const { settings, isVeto, t } = params;
+        const { settings, isVeto, t, realTimeTotalSupply = '0' } = params;
```
tyhonchik, line 108: `realTimeTotalSupply = '0'` so during load we show `participationNoToken` (% only), then snap to the version with token value. Could we surface `isLoading` and render a skeleton/`—` until data lands?

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
```diff
@@ -70,7 +72,7 @@ class LockToVoteSettingsUtils {
         );
 
         const minParticipationToken = Math.round(
```
tyhonchik: worth extracting a shared helper as commented for lockToVoteProcessBodyField.tsx

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
```diff
@@ -70,7 +72,7 @@ class LockToVoteSettingsUtils {
...
-            (Number(processedTotalSupply) * parsedMinParticipation) / 100,
+            (Number(realTimeTotalSupply) * parsedMinParticipation) / 100,
```
tyhonchik: 18-decimal LTV tokens can have `totalSupply > 2^53`. Pre-existing pattern, but now we're computing off a live supply

maybe it's good moment to switch to bigint: `(totalSupply * bigIntUtils.safeParse(minParticipation)) / 100n`?

src/modules/governance/api/governanceService/utils/collectTokenAddresses/collectTokenAddresses.ts
```diff
@@ -0,0 +1,35 @@
...
+import { lockToVoteProposalUtils } from '@/plugins/lockToVotePlugin/utils/lockToVoteProposalUtils';
+import { sppProposalUtils } from '@/plugins/sppPlugin/utils/sppProposalUtils';
+import type { IProposal } from '../../domain';
+
+/**
+ * Walks a proposal (and, when present, its SPP stages and sub-proposals) and returns the unique
+ * lowercased token addresses for plugins/bodies whose status math requires a live `totalSupply`
+ * read. Today only lock-to-vote bodies qualify.
+ */
+export const collectTokenAddresses = (proposal: IProposal): Hex[] => {
```
tyhonchik, line 11: imo would be good to validate addresses and add proper typings to avoid casting in this util

src/modules/governance/api/governanceService/utils/fetchTokensTotalSupply/fetchTokensTotalSupply.ts
```diff
@@ -0,0 +1,85 @@
...
+): Promise<bigint[]> => {
+    const rpcUrl = resolveServerRpcUrl(network);
+
+    if (rpcUrl == null) {
+        throw new Error(
+            `fetchTokensTotalSupply: no server-side RPC endpoint configured for network ${network}`,
+        );
+    }
+
+    const client = createPublicClient({
```
tyhonchik, line 76: `createPublicClient` is rebuilt per request
looks like we can memoize it by network

tyhonchik (commented):
Hey, nice PR
A few comments below, no blockers

src/plugins/lockToVotePlugin/components/lockToVoteProposalVotingSummary/lockToVoteProposalVotingSummary.tsx
milosh86: yeah, I already evaluated this exact question! The issue here is that there is no good fallback to the real-time totalSupply. Historical supply or anything else would simply provide wrong statuses and other information about the proposal.

I considered an approach to strip all information derived from totalSupply in the case of missing one, but that would be a major task. Let's track and see if we ever get this issue, and we can act upon it if needed.

src/modules/governance/api/governanceService/governanceService.ts
milosh86: This should be ok in general, we don't have a use-case with multi-network operations. Anyway, I will add here invariant to at least be obvious what are the expectations, just in case we go that direction one day. It's a good point that invariant is invisible.

src/modules/application/utils/proxyRpcUtils/resolveServerRpcUrl.ts
milosh86: I'll switch to NEXT_RUNTIME here

src/plugins/lockToVotePlugin/components/lockToVoteProcessBodyField/lockToVoteProcessBodyField.tsx
milosh86: created reusable util

src/plugins/lockToVotePlugin/components/lockToVoteProcessBodyField/lockToVoteProcessBodyField.tsx
milosh86: makes sense

src/plugins/lockToVotePlugin/hooks/useLockToVoteGovernanceSettings/useLockToVoteGovernanceSettings.ts
milosh86: done

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
milosh86: I was thinking about this, but there are places like `normalizeChangeSettingsAction` where we don't fetch totalSupply at all.

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
milosh86: done

src/modules/governance/api/governanceService/utils/collectTokenAddresses/collectTokenAddresses.ts
milosh86: casting here is only because the type of address on token object is `string`.... it only operates on `settings.token.address`, I think additional validation would be excessive here. If we don't trust data from server, we should consider validating responses in service layer, before passing it into the app logic.

src/modules/governance/api/governanceService/utils/fetchTokensTotalSupply/fetchTokensTotalSupply.ts
milosh86: tbh, not sure it's worth the complexity ... let me check what `createPublicClient` exactly does

milosh86: yeah, it just move around plain objects, I would leave it as is. WDYT?

src/plugins/lockToVotePlugin/utils/lockToVoteSettingsUtils/lockToVoteSettingsUtils.tsx
milosh86: I can actually show skeletons in `LockToVoteProcessBodyField`, and in general this is the only place where the "snap" is visible.

src/plugins/lockToVotePlugin/utils/lockToVoteProposalUtils/lockToVoteProposalUtils.ts
```diff
@@ -132,12 +141,21 @@ class LockToVoteProposalUtils {
...
+        plugin.interfaceType === PluginInterfaceType.LOCK_TO_VOTE;
+
     getProposalTokenTotalSupply = (proposal: ILockToVoteProposal) => {
-        const { historicalTotalSupply, token } = proposal.settings;
-        // Fallback to the token total-supply as some plugins (e.g. lock-to-vote) do not use snapshot votes.
-        const totalSupply = historicalTotalSupply ?? token.totalSupply;
+        const tokenAddress = proposal.settings.token.address.toLowerCase();
 
-        return totalSupply;
+        return proposal.tokensTotalSupply[tokenAddress];
```
thekidnamedkd, line 158: I think this changes the proposal math in a risky way. We used to rely on the proposal’s snapshot supply via `historicalTotalSupply`, but this now always pulls from the live `tokensTotalSupply` map. If supply changes after proposal creation, the outcome and participation checks can shift after the fact.

src/modules/governance/api/governanceService/governanceService.ts
```diff
@@ -74,18 +79,68 @@ class GovernanceService extends AragonBackendService {
...
+            'GovernanceService.getProposalList: all proposals must share the same network',
+        );
+
+        const tokensTotalSupply =
+            addresses.length === 0 || result.data.length === 0
+                ? {}
+                : await fetchTokensTotalSupply(
+                      result.data[0].network,
+                      addresses,
+                  );
```
thekidnamedkd, line 97: This also looks a bit fragile now that proposal fetches await `fetchTokensTotalSupply(...)`. If that realtime RPC call fails or times out, we end up rejecting the whole request and the proposal page can error, even though we already had enough supply data in the existing payload to render before.

src/plugins/lockToVotePlugin/utils/lockToVoteProposalUtils/lockToVoteProposalUtils.ts
milosh86: As far as I understand, that's how LTV works in fact, all proposal math is based on the current up-to-date totalSupply, it doesn't use historicalTotalSupply. The bug was related to using historicalTotalSupply instead of current total supply.

src/modules/governance/api/governanceService/governanceService.ts
milosh86: connected to the above comment ... if proposal calculations are based on the real-time total supply, then there is no viable fallback and only reasonable thing is to treat RPC error the same way as BE error.

> even though we already had enough supply data in the existing payload to render before.

as mentioned, we can't treat this as a fallback, because it's simply wrong source of data in general, even though at some timestamps it could be valid.

thekidnamedkd (approved):
LGTM⚡️

PR #1137, opened 2026-05-18
fix(APP-630): Allow fractional gauge votes via relative-weight dialog

thekidnamedkd:
## Summary

- Drops the sum-to-100 UI rule that caused decimal weight inputs to silently disable Submit.
- Replaces percentages with integer relative weights stored as `bigint` via `parseUnits` / `formatUnits`. No JS floats in the value path.
- Aligns with the `AddressGaugeVoter` contract: weights are relative `uint256`, normalized on-chain; sole invariant is `Σweights > 0`.
- Default-equal weights on first open. Equalize → all `1`. Reset → all `0`.
- Per-row dynamic share helper, always visible.
- Footer reads "100% of your voting power will be allocated as follows."
- Weight input capped at 1000 (UX-only; contract is scale-invariant).
- Displayed shares are rounded via Hamilton's largest-remainder so they always sum to exactly 100%. Previously two rows showing `99.90% / 0.09%` summed to `99.99%` while the footer claimed `100%`. Now: `99.90% / 0.10%`.
- 20 unit tests covering the dialog state model, per-row display math, and the apportionment-rounding edge cases.
- Gauge voter parent now waits for both wallet connection AND voting data to load before opening the vote dialog (via a `voteRequested` state + effect). Previously, a disconnect-state first click opened the dialog with stale `0` voting power because the data fetch hadn't completed when the wallet-guard fired its `onSuccess`.

Closes APP-630.

**Preview path:** `/dao/ethereum-sepolia/0x4ec795c6F2a67634C2E8e7Fd9887dc83283D059b/gauges`

## Why weights, not percentages

The original implementation used percentage inputs constrained to sum to 100. Two structural problems caused the ticket bug:

1. **Sum-to-100 forces rebalancing.** Every edit on one row implies a change to the others to keep the total at 100. Either auto-rebalance (clobbers user intent the user didn't ask for) or block submit until they manually patch (the silent-disable bug). With weights, rows are independent — `A=1, B=3` is `25% / 75%`; bumping A to `2` becomes `40% / 60%` without touching B.
2. **Precision loss on the round-trip.** Contract takes relative weights. Expressing "1/3 to each of 3 gauges" via percentages forces `33.33 / 33.33 / 33.34` and round-trip mapping back to weights drops precision — exactly the "fractional gauge votes cannot be submitted" symptom. With weights as the source of truth, `1 / 1 / 1` submits as `1 / 1 / 1` exactly.

### Why not a cosmetic `%` mask over a weight input

Considered: keep input semantics as weight but render a trailing `%` on the field. Rejected — `<input>` carries semantics (value, range, label, type), and decoration that contradicts those semantics is an a11y bug, not styling. Screen readers, voice control, autofill and validators all announce the underlying value+range, so a `%` suffix on a `0–1000` weight input misrepresents the field in every assistive context. Current implementation keeps the input weight-only and shows the derived percentage as a separate adjacent readout.

### Why not rebalance percentages internally

Considered: percentage input + transparent renormalization underneath. Rejected — you still have to *interpret* every edit (proportional rebalance? lock siblings? let total drift?). Each policy surprises users in different cases. Independent weights with a derived-percent display sidesteps the interpretation problem entirely.

### Apportionment rounding (Hamilton's largest remainder)

Displayed parts rounded independently aren't guaranteed to sum to the whole. With weights `[1000, 1]`, naive rounding gives `99.90 / 0.09 = 99.99%`. Hamilton's method: floor each share at display precision, distribute the leftover quanta to rows with the largest fractional remainder. Same case becomes `99.90 / 0.10 = 100.00%`. Implemented in `gaugeVoterVoteDialogContent` so all row shares are computed together; items render the supplied `displayShare` rather than computing their own.

## Recommended QA

- [ ] Open the vote dialog with 2–3 gauges (already connected). Default state shows `1` per gauge with equal shares summing to 100% (will read `33.34% / 33.33% / 33.33%` for three gauges — sum is exact).
- [ ] Type a decimal (e.g. `0.5`) in a weight input — Submit remains enabled (no silent failure).
- [ ] Set weights `[5, 3, 2]`. Verify shares display `50% / 30% / 20%`.
- [ ] Set weights `[1000, 1]`. Verify shares display `99.90% / 0.10%` (sum to exactly 100%, not 99.99%).
- [ ] Click Equalize. All weights → `1`.
- [ ] Click Reset. All weights → `0`. Submit disables.
- [ ] Type a weight above 1000. Cap holds.
- [ ] Reopen the dialog after a successful vote. Inputs pre-populate from prior on-chain allocation.
- [ ] **Disconnect-state first click**: connect wallet via the "Vote on selected" prompt, then verify the dialog opens with your actual voting power (not `0` MERC).

## Related

- **APP-712** — separate defensive improvement to `useConnectedWalletGuard` (safe-by-default `onSuccess` pattern). Not required to close APP-630; tracked independently.

## Type of Change

## Developer Checklist

## Review Checklist


milosh86 (approved):
Great work really! 🥇

PR #1143, opened 2026-05-20
feat(APP-604): Add new SelectAddress component for asset selection

milosh86:
# Description

Introduces a new `AssetAddressSelect` component replacing the existing `AssetSelectionDialog` flow in `AssetInput`. Users can now pick a token from the DAO's asset list, search by name/symbol/address, or add a custom token by entering its address (resolved on the fly via RPC).

Linear: [APP-604](https://linear.app/aragon/issue/APP-604/new-selectaddress-component)

### Highlights
- New `AssetAddressSelect` component with paginated list, search, and "add by address" flow
- Subcomponents: `AssetAddressSelectItem`, `AssetAddressSelectAddAddressView`, `AssetAddressSelectAddButton`, `AssetAddressSelectBackButton`, `AssetAddressSelectEmptyState`
- Search by name, symbol, or address; auto-resolves addresses immediately when the query is an address
- `AssetInput` updated to use the new selector and to flip the max-amount layout per design
- Token symbol/logo fallbacks tightened in `AssetInputToken`
- Bumps `@aragon/gov-ui-kit` to use the new `DataListItem` select variant

## Type of Change

## Developer Checklist:

## Review Checklist:


src/modules/finance/components/assetAddressSelect/assetAddressSelectAddAddressView.tsx
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    const asset: IAsset | undefined =
+        resolvedAddress != null
+            ? {
+                  amount: '0',
+                  token: {
+                      address: resolvedAddress,
+                      network,
+                      name: tokenData?.name ?? '',
```
milosh86: default to empty string or to some default values like `Unknown`?

src/modules/finance/components/assetAddressSelect/assetAddressSelectAddButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+            className={classNames(
+                'group flex w-full cursor-pointer items-center gap-3 rounded-xl bg-transparent px-4 py-3 text-base text-primary-400 leading-tight transition-colors',
+                // Default bg is transparent; hover/focus use primary-400 @ 4%, active @ 8%
+                'focus-ring-primary hover:bg-primary-400/4 active:bg-primary-400/8',
+                'md:gap-4 md:px-6 md:py-5 md:text-lg',
+                className,
+            )}
+            type={type}
+            {...otherProps}
+        >
```
thekidnamedkd: there's a new variant update coming? or the designs branch here from govkit?

src/modules/finance/components/assetAddressSelect/assetAddressSelectEmptyState.ts
```diff
@@ -0,0 +1,35 @@
+import { IconType } from '@aragon/gov-ui-kit';
```
thekidnamedkd: I don't remember this pattern. It's sort of like hook/util as component? Naming and location seems off maybe.

src/shared/utils/pluginTransactionUtils/pluginTransactionUtils.ts
```diff
@@ -119,14 +119,17 @@ class PluginTransactionUtils {
...
+            !addressUtils.isAddress(pluginSetupRepo)
+        ) {
             throw new Error(
                 `Plugin repository address not found for interface type: ${plugin.interfaceType}`,
             );
         }
 
         const pluginSetupRef = {
-            pluginSetupRepo,
+            pluginSetupRepo: pluginSetupRepo as Hex,
```
thekidnamedkd, line 132: Do we need to cast?

src/modules/finance/components/assetAddressSelect/assetAddressSelectItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    const processedEntityUrl = onAssetClick != null ? undefined : entityUrl;
+    const processedTarget = onAssetClick != null ? undefined : '_blank';
+
+    const hasPrice = Number(token.priceUsd) > 0;
+    const fiatPrice = hasPrice ? token.priceUsd : '0';
+
+    return (
+        <AssetDataListItemStructure
+            amount={amount ?? '0'}
```
thekidnamedkd, line 40: Does this handle nullish internally?

thekidnamedkd (commented):
Looking good. Couple questions. 🧑‍🚀

src/shared/utils/pluginTransactionUtils/pluginTransactionUtils.ts
milosh86: this was strange, I don't know why I've got this TS error, I didn't change anything related .... I'll double check, maybe just some TS cache issue

src/modules/finance/components/assetAddressSelect/assetAddressSelectEmptyState.ts
milosh86: right, I was on the fence when I saw it and I just let it go, but you are right, no need to have a different pattern for the same thing we already have in other places.

src/modules/finance/components/assetAddressSelect/assetAddressSelectAddButton.tsx
milosh86: I didn't want to overload the DataListItem with another variant just for this small component. I will see if I can create a new component for this one and also for back button which uses different color and icon, but the structure is the same

src/modules/finance/components/assetAddressSelect/assetAddressSelectItem.tsx
milosh86: I'll double check, I think this was needed, otherwise it shows NaN

src/modules/finance/components/assetAddressSelect/assetAddressSelectAddButton.tsx
milosh86: I added new component to ui-kit which is used as base for both Add and Back buttons

src/modules/finance/components/assetAddressSelect/assetAddressSelectEmptyState.ts
milosh86: .... I just inlined empty state as it's passed as a prop

src/shared/utils/pluginTransactionUtils/pluginTransactionUtils.ts
milosh86: could you please check if you get this TS error in main? I do, and I don't know how we got around it before...

src/modules/finance/components/assetAddressSelect/assetAddressSelectItem.tsx
milosh86: yeah, it's not handled internally, it expects `string | number`

src/modules/finance/components/assetAddressSelect/assetAddressSelectAddAddressView.tsx
```diff
@@ -0,0 +1,112 @@
...
+                      symbol: tokenData?.symbol ?? '',
```
thekidnamedkd: When `useToken` fails to resolve (non-ERC20, RPC error), this row still gets `decimals: 0` and stays selectable. Downstream `transferAssetAction.tsx:110` reads `decimals ?? 18` — but `0` is a real number, so the fallback never fires and `parseUnits(amount, 0)` throws on any fractional amount. Either block selection until `tokenData != null`, or fall back to `18` here (matching `useGovernanceToken`). Given the Cryptex/Katana context I'd lean toward the stricter gate.

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    const filteredAssets = useMemo(() => {
+        if (!(hasSearch && searchValue)) {
+            return assetList;
+        }
+
+        const lowerSearch = searchValue.toLowerCase();
+        const searchIsAddress = addressUtils.isAddress(searchValue);
+
+        return assetList.filter(({ token }) => {
```
thekidnamedkd, line 87: This filter only sees already-loaded pages, and pagination is hidden once `searchValue` is set. So on a treasury larger than `pageSize`, a held token on page 2+ silently fails to match and the auto-pivot below then routes the user into Add-by-address as if the DAO doesn't hold it. Server-side search via `useAssetList` is probably the right fix — happy to follow up if that's separate work.

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
```diff
@@ -0,0 +1,173 @@
...
+            return;
+        }
+        if (filteredAssets.length > 0) {
+            return;
+        }
+
+        setPendingAddress(searchValue as Hex);
+        setSearchValue(undefined);
+        setMode('addAddress');
+    }, [mode, searchValue, filteredAssets.length]);
```
thekidnamedkd, line 111: Tiny thing: the dependency here is `filteredAssets.length` rather than `filteredAssets`. Could re-trigger the effect in a loop. Might be worth using the array reference itself, or extracting a stable boolean.

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
```diff
@@ -0,0 +1,173 @@
...
+            )}
+            <DataListContainer
+                emptyState={emptyState}
+                errorState={errorState}
+                SkeletonElement={AssetDataListItem.Skeleton}
+            >
+                {filteredAssets.map((asset, index) => (
+                    <AssetAddressSelectItem
+                        asset={asset}
+                        key={`${asset.token.address}-${index.toString()}`}
```
thekidnamedkd: The token address is already unique within a treasury, so appending the array index here defeats React reconciliation when the list reorders or filters and causes needless remounts. Just `key={asset.token.address}` should be enough.

src/modules/finance/components/assetAddressSelect/assetAddressSelectItem.tsx
```diff
@@ -0,0 +1,53 @@
...
+            fiatPrice={fiatPrice}
+            hideValue={false}
+            href={processedEntityUrl}
+            key={token.address}
+            logoSrc={token.logo}
+            name={token.name || 'Unknown'}
+            onClick={() => onAssetClick?.(asset)}
+            symbol={token.symbol || 'UNKNOWN'}
```
thekidnamedkd: `Unknown` and `UNKNOWN` strings should probably go through `useTranslations` 

One thing to note: `useAssetListData` already maps blank API responses to `Unknown` / `UNKNOWN` upstream, so these fallbacks really only fire for the on-the-fly resolved row, which ties back to the decimals comment on the add-address view.

thekidnamedkd (commented):
Did a closer pass — overall this is shaping up nicely. A few things worth a look, the biggest being the decimals fallback on the Unknown asset row, which I think will throw downstream in the transfer flow.

src/modules/finance/components/assetAddressSelect/assetAddressSelectAddAddressView.tsx
milosh86: yeah, an asset could be without name and symbol, but we should only accept it if `decimals` property exist. I'll update the logic, not to allow selection of an asset without strict decimals. I would rather not fallback to 18, that's not safe practice as we saw.

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
milosh86: yes, that's how the current version works, I just adopted the same approach ... do you remember if there is a reason to do so?

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
milosh86: hm, not sure about this, only `filteredAssets.length` is used inside the effect and it's a simple number type, so it won't change across renders. What would be the case when it could trigger a loop rendering?

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
milosh86: true!

src/modules/finance/components/assetAddressSelect/assetAddressSelectItem.tsx
milosh86: I didn't notice that useAssetListData sets those fallbacks already ... I'll just remove the logic from here

src/modules/finance/components/assetAddressSelect/assetAddressSelect.tsx
milosh86: double checked, current version works like this, so let's leave this as a follow-up ticket if needed.

thekidnamedkd (approved):
LGTM 🔥

PR #1162, opened 2026-06-01
fix(APP-789): Keep input focus when editing decoded proposal action fields

milosh86:
# Description

Editing any parameter of a proposal action in the **Decoded** view lost input focus after every keystroke — you had to click back into the field to type the next character.

**Root cause:** the action item remounted on every keystroke. The gov-ui-kit decoder re-encodes parameters and writes the result to `actions.N.data` via `setValue` on each keystroke (by design — the form must always hold the encoded calldata). In react-hook-form, calling `setValue` on a *child* of a field array emits the array subject, which makes `useFieldArray` regenerate every `field.id`. Since the rendered item was keyed by `field.id`, the whole item unmounted/remounted and destroyed the focused `<input>`.

**Fix:** give each action a form-owned, stable `fieldId` (assigned once in `handleAddAction`) and use it as the React `key`/`value` instead of RHF's regenerating `field.id`. `fieldId` lives in the form values, so it survives the field-array id regeneration. RHF's default `keyName` is left untouched to avoid coupling to library internals.

`fieldId` is whitelisted out of both the on-chain submit (`actionToTransactionRequest` picks only `{ to, value, data }`) and the JSON export (`exportActionsToJSON` maps only `{ to, value, data }`), so it never leaks into payloads.

The gov-ui-kit side is fixed and verified separately; this PR is the consumer-side identity-stability fix.

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (approved):
lgtm 👍

PR #1169, opened 2026-06-04
perf(APP-809): Improve image caching, ISR revalidation and dialog code splitting

milosh86:
# Description

Set of performance optimisations around static assets and bundle size:

- **IPFS image caching**: set `minimumCacheTTL` to 30 days on the Next.js image optimizer — IPFS content is CID-addressed and immutable, kept below 1 year since the TTL is global and shared with mutable hosts (CMS logos, token icons)
- **ISR revalidation**: add `revalidate` to the explore page (5 min) and DAO page layout (1 min) so they are served statically and regenerated in the background
- **Dynamic dialog imports**: import all module dialogs (`dispatchDialog`, `routerSelectorDialog`, `createDaoDetailsDialog`, `setupBodyDialog`, `assetSelectionDialog`, `executeDialog`) via `next/dynamic` so they are split into separate chunks and loaded on demand
- **Font cleanup**: remove redundant `.woff` fallbacks for TWK Everett (the `.woff2` variants remain)
- Small `ipfsUtils` adjustment

## Type of Change

## Developer Checklist:

## Review Checklist:


src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx
```diff
@@ -1,5 +1,8 @@
 import { LayoutDao } from '@/modules/application/components/layouts/layoutDao';
 import { applicationMetadataUtils } from '@/modules/application/utils/applicationMetadataUtils';
 
+export const revalidate = 60; // ISR: regenerate every minute
```
Comment (tyhonchik) [inline src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx]: I don't think revalidate actually enables ISR here. The root layout (layoutRoot.tsx) calls `headers()`, which makes the whole route tree dynamic. Also our backend calls use cache: `'no-store'` by default (httpService.ts). So these pages still render SSR on every request, and this config has no effect.

Two options:
a) Remove these two lines for now, since they don't do anything and may confuse the next reader + create ticket fro the proper ISR
Or
b) do the full work to make ISR real: move `headers()` (wagmi cookie + feature flags) out of the root layout, switch the DAO/explore services to `force-cache` with `next.revalidate`, and maybe add `generateStaticParams`.
What do you think?

src/app/page.tsx
```diff
@@ -1,3 +1,5 @@
 import { ExploreDaosPage } from '@/modules/explore/pages/exploreDaosPage';
 
+export const revalidate = 300; // ISR: regenerate every 5 minutes
```
tyhonchik: the same comment about ISR

next.config.mjs
```diff
@@ -127,6 +127,9 @@ const nextConfig = {
         ];
     },
     images: {
+        // 30 days — IPFS content is immutable (CID-addressed); kept below 1 year because the TTL is global
+        // and mutable hosts (CMS logos, token icons) share the optimizer with no invalidation mechanism.
+        minimumCacheTTL: 2_592_000,
```
tyhonchik: Good idea for IPFS images. One thing to double check: this 30-day value is a minimum, so it also applies to mutable images (CMS logos, token icons) even if server sends a short cache time, and there is no way to invalidate them. If a token icon changes, users may see the old one for up to 30 days. Are we ok with that?

If not, maybe a smaller value (e.g. 7 days) or a cache-busting query param on mutable URLs would be safer.

src/modules/governance/dialogs/executeDialog/index.ts
```diff
@@ -1,5 +1,10 @@
-export {
-    ExecuteDialog,
-    type IExecuteDialogParams,
-    type IExecuteDialogProps,
+import dynamic from 'next/dynamic';
+
+export const ExecuteDialog = dynamic(() =>
+    import('./executeDialog').then((mod) => mod.ExecuteDialog),
```
tyhonchik, line 4: Small UX note: with no loading option, on a slow connection the modal frame shows empty for a moment while the chunk loads. Not blocking, just maybe a small spinner would feel nicer. Our other dialogs also skip it, so up to you.

src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx
Comment (tyhonchik) [inline src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx:1]: We now have some dialogs dynamic and some static, with no written rule for when to use dynamic. Could we add a short note somewhere (or finish converting the rest) so the next person doesn't have to guess?

src/modules/finance/dialogs/assetSelectionDialog/index.ts
```diff
@@ -1,6 +1,6 @@
 import dynamic from 'next/dynamic';
 
-export const PublishProposalDialog = dynamic(() =>
+export const AssetSelectionDialog = dynamic(() =>
```
tyhonchik, line 3: good catch!

tyhonchik (changes requested):
Nice set of optimizations! The AssetSelectionDialog fix is a ✨. I have one bigger question about the ISR part and a small note about image caching.

jjavieralv (dismissed):
LGTM

src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx
Comment (milosh86) [reply src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx]: good catch, and I kinda validated this by checking Vercel logs where I couldn't see any improvement on ISR.

I'll just remove it for now.

next.config.mjs
milosh86: yeah, I was thinking about this and currently we only use image for IPFS avatars and token logos, so I couldn't see a case where we change an image and the name stays the same. Also, as you mentioned, we can add a query param to invalidate cache if we get such a case.

WDYT? .. maybe I'll reduce to 7 days as you suggested anyway, let's start with that value.

src/modules/governance/dialogs/executeDialog/index.ts
milosh86: good catch ... this would be more like a new pattern that should be applied to all dialogs, so let's have this as a separate ticket

src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx
Comment (milosh86) [reply src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx:1]: hm, all dialogs should be dynamic, in this PR I just "fixed" remaining static dialogs.
I'll add agents rule.

Comment (milosh86) [reply src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx:1]: are there any static dialogs left that you see?

next.config.mjs
tyhonchik: agree, let's start with simpler approach

src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx
Comment (tyhonchik) [reply src/app/dao/[network]/[addressOrEns]/(daoPage)/layout.tsx:1]: ah, I double-checked, my mistake
all good here, I checked dialogs on the old branch before

tyhonchik (approved):
LGTM🚀

PR #1171, opened 2026-06-08
feat(APP-810): Add action simulation to the direct execute actions flow

milosh86:
# Description

Adds optional Tenderly action simulation to the **Direct Execute Actions** flow, mirroring the affordance already present in the Create Proposal flow: on the actions step the **Execute** button becomes a dropdown with **Simulate actions** / **Skip simulation**.

The simulation logic is shared rather than copied. Both flows already use the same action shape, the same form provider, and the same action-preparation util, so the only per-flow differences are the executor address and the form id.

**Key changes**
- New shared hook `useSimulateActionsDropdown({ daoId, from, formId })` — validates the form, prepares actions, opens the `SIMULATE_ACTIONS` dialog, and returns the dropdown items gated on `actions.length > 0 && tenderlySupport && from != null`. Both flows consume it; ~50 lines of inline logic removed from `createProposalPageClientSteps`.
- Renamed the simulation address param `pluginAddress` → `from` across the API, service URL template, and dialog, since it is semantically just the `from` sender.
- The execute flow simulates from the **connected wallet** — the account that actually calls `DAO.execute` (the analog of the proposal flow simulating from the plugin). `from = DAO` would be wrong: the DAO calling its own `execute` reverts.
- Moved the shared `simulate` / `skipSimulation` i18n labels into the `simulateActionsDialog` namespace.

**Note for reviewers / backend dependency:** the current backend `simulate` endpoint resolves the DAO from a *plugin* lookup and pins `to`/`from` to that plugin, so it does not yet serve the plugin-less direct-execute case. A DAO-scoped endpoint variant is required for the simulation to succeed end-to-end; the frontend here passes the semantically correct `from` (connected wallet) in anticipation of that.

## Type of Change

## Developer Checklist:

## Review Checklist:


src/shared/components/networkSwitchAlert/networkSwitchAlert.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useTranslations } from '../translationsProvider';
+
+export interface INetworkSwitchAlertProps {
+    /** Whether the wallet's current chain differs from the required chain. */
+    isCrossNetworkTransaction: boolean;
+    /** Human-readable name of the required network. */
+    networkName?: string;
+}
+
+export const NetworkSwitchAlert: React.FC<INetworkSwitchAlertProps> = (
```
tyhonchik, line 11: thanks for creating new component instead of inline switch alert that I brought recently

src/modules/governance/dialogs/executeActionsDialog/executeActionsDialog.tsx
```diff
@@ -54,6 +56,13 @@ export const ExecuteActionsDialog: React.FC<IExecuteActionsDialogProps> = (
     const { close } = useDialogContext();
     const { setIsBlocked } = useBlockNavigationContext();
     const { data: dao } = useDao({ urlParams: { id: daoId } });
+    const {
+        requiredChainId,
+        isCrossNetworkTransaction,
+        networkName,
+        switchChainStatus,
+        withNetworkSwitch,
+    } = useNetworkSwitch({ daoId });
```
tyhonchik: `useNetworkSwitch` returns `isLoading`, but we don't consume it here.

While the DAO is still resolving `sendTransaction` may go out with `chainId: undefined`, bypassing the chain-pin guard.

If the submit can be reached before the DAO loads, let's pull `isLoading` and fold it into the submit's `disabled` or `isLoading` state.

src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -65,23 +57,21 @@ export const TransactionDialog = <TCustomStepId extends string>(
...
-        network,
-    });
-
-    const isCrossNetworkTransaction =
-        chainId != null &&
-        requiredChainId != null &&
-        chainId !== requiredChainId;
-    const transactionNetworkName = networkDefinitions[network]?.name;
+    const { address } = useWalletAccount();
+    const { buildEntityUrl } = useDaoChain({ network });
```
tyhonchik, line 67: `useNetworkSwitch({ network })` already calls `useDaoChain` internally, so we now run `useDaoChain` (and `useBlockExplorer`) twice per render with the same network.

Let's drop the duplicate — e.g. have useNetworkSwitch also return `buildEntityUrl` and `networkDefinition`, or keep the single `useDaoChain` call and pass its result down.

src/modules/governance/hooks/useSimulateActionsDropdown/useSimulateActionsDropdown.ts
```diff
@@ -0,0 +1,104 @@
...
+    const { prepareActions } = useCreateProposalFormContext();
+
+    const { network, address: daoAddress } = daoUtils.parseDaoId(daoId);
+    const { tenderlySupport } = networkDefinitions[network];
+    const watchedActions =
+        useWatch<Record<string, ICreateProposalFormData['actions']>>({
+            name: 'actions',
+        }) ?? [];
+
+    const handleSimulate = async () => {
```
tyhonchik: `handleSimulate` and the returned array are rebuilt on every render, so `nextDropdownItems` is a new reference each time.

If `WizardPage.Step` is render-sensitive to its props this creates re-renders. Suggest wrapping `handleSimulate` in `useCallback` and the items array in `useMemo`.

src/shared/hooks/useNetworkSwitch/useNetworkSwitch.api.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { TransactionStatusState } from '@/shared/components/transactionStatus';
+
+export interface IUseNetworkSwitchParams {
+    /** Network to resolve the required chain from. */
+    network?: Network;
+    /** DAO ID to derive the required chain from (fetches DAO if network/chainId not provided). */
+    daoId?: string;
+    /** Required chain ID to use directly (bypasses DAO and network resolution). */
+    chainId?: number;
+}
```
tyhonchik, line 7: `network`, `daoId`, and `chainId` are all optional with no validation.

If none is passed, `requiredChainId` is `undefined` and the hook silently reports "no switch needed", which is easy to misuse.

Let's document the resolution precedence or assert that at least one source is provided.

tyhonchik (commented):
good PR
just left couple comments regarding efficiency and loading state

src/shared/components/transactionDialog/transactionDialog.tsx
milosh86: tbh, both approaches would break abstraction  ... but, I just realised this comment is related to the last one about params format. This is a bit weird, I'll update `useNetworkSwitch` to accept `network` only and in that case there is no "async" operation at all, so no perf issue too.

src/modules/governance/dialogs/executeActionsDialog/executeActionsDialog.tsx
milosh86: In the end refactored `useSwitchNetwork` and removed `isLoading` ... it's "sync" hook now.

src/shared/hooks/useNetworkSwitch/useNetworkSwitch.api.ts
milosh86: Refactored params to accept only `network`

src/shared/components/transactionDialog/transactionDialog.tsx
milosh86: Refactored useNetworkSwitch not to call `useDaoChain`.

tyhonchik (approved):
LGTM! 🎶

PR #1175, opened 2026-06-09
feat(APP-795): Integrate Aragon subdomain and implement Aragon profile renaming

milosh86:
# Description

- Propose a minimal approach for `aragon-subdomain` integration (see the section below)
- Implement Aragon name changing 

**Integration approach:**

- We use the same "Service" abstraction as we already use in the App for interacting with backend and RPC
- Each service method has its own API route
  - This is different to the existing backend and RPC based services which have single API route acting as a proxy!
  - We need to implement common operations like response parsing, error handling and monitoring in a reusable way
- We have shared AragonSubdomainServiceClient used in HTTP client path (extended by specific client services)
- We have AragonSubdomainServiceServer used in the API routes and in the service server path
- Specific service implementation, e.g. memberProfileService
  - Service has a client and a server part
  - Domain interfaces should come straight from the `aragon-subdomain`, instead of re-implementing the same interfaces
  - `api.ts` file implements all "params" interfaces
  - Even though server path uses domain controller which is not HTTP interface, we should probably keep param interfaces in the same format in all cases... to be confirmed
  - Client part of the service is identical to the existing backend services, except it extends `AragonSubdomainServiceClient`
  - Server part of the service has identical interface as the client part, but internally, it uses shared `aragonSubdomainServiceBackend`, which is basically instantiated AragonDomainController
- Tanstack queries
  - Queries have identical structure to existing backend queries, except they are only intended for the client side!
  - For the server component calls, using `queryClient.fetchQuery(queryOptions())`, we have `.server.ts` query file which exports only a function for generating server component query options. It reuses client function completely plus it adds new `queryFn`
  - For the sake of caching, it is important to use the same query key in both client and server case!

## Type of Change

## Developer Checklist:


src/modules/ens/hooks/useAllEnsRecords.ts
```diff
@@ -0,0 +1,37 @@
+export function useAllEnsRecords(ensName: string) {
```
milosh86: please ignore this one

src/modules/governance/pages/memberProfileDemoPage/memberProfileDemoPage.tsx
```diff
@@ -0,0 +1,40 @@
...
+}
+
+export interface IMemberProfileDemoPageProps {
+    /**
+     * Demo page route parameters.
+     */
+    params: Promise<IMemberProfileDemoPageParams>;
+}
+
+export const MemberProfileDemoPage: React.FC<
```
milosh86: TO be removed before merging!

src/modules/application/dialogs/aragonProfileRenameDialog/aragonProfileRenameDialog.tsx
```diff
@@ -54,16 +65,38 @@ export const AragonProfileRenameDialog: React.FC<
...
+
         open(ApplicationDialogId.ARAGON_PROFILE_RENAME_TRANSACTION, {
             stack: true,
-            params: { subdomain },
+            params: {
+                subdomain,
+                records: {
+                    textRecords: allTextRecords ?? [],
+                    addr: address,
+                    contenthash: '0x',
```
thekidnamedkd: Think we've got a data-loss issue here 👀 — `contenthash: '0x'` wipes any existing contenthash, and `addr` is the connected wallet rather than the name's current `addr` record. On top of that `getEnsTextRecords` only pulls text records, so addr/contenthash never get read from the profile at all.

Since the AC is "move **all** records", we should fetch the full `Records` (text + addr + contenthash) for the current name and pass them through unchanged. Worth testing against a profile that actually has a contenthash set so we don't silently nuke it on rename.

src/shared/api/aragonSubdomainService/aragonSubdomainService.backend.ts
```diff
@@ -0,0 +1,13 @@
+import 'server-only';
+import { AragonSubdomain, EnvioClient } from '@aragon/aragon-subdomain';
+
+const endpoint = process.env.NEXT_SECRET_ENVIO_GRAPHQL_ENDPOINT;
+const apiToken = process.env.NEXT_SECRET_ENVIO_API_TOKEN;
+
+if (endpoint == null) {
+    throw new Error('Envio endpoint is not set');
+}
```
thekidnamedkd: This throws at import time, before the route's try/catch can catch it — so a missing endpoint surfaces as an opaque import crash instead of a clean 500. Fine if we trust deploy config, but can we double-check both new secrets are set in every preview/prod env where the profile dialog renders?

Also `apiToken` has no null check before it goes into `EnvioClient` — intentional (token's optional) or a miss?

src/modules/application/api/memberProfileService/memberProfileService.client.ts
```diff
@@ -0,0 +1,20 @@
+import { AragonSubdomainServiceClient } from '@/shared/api/aragonSubdomainService';
+import type { IMemberProfileTextRecord } from './domain/IMemberProfileTextRecord';
+import type { IGetEnsTextRecordsParams } from './memberProfileService.api';
+
+class MemberProfileServiceClient extends AragonSubdomainServiceClient {
+    private urls = {
+        getMemberProfileTextRecords: '/member-profile/:name/records',
+    };
+
+    getEnsTextRecords = async (params: IGetEnsTextRecordsParams) => {
```
thekidnamedkd, line 10: The new service + both dialogs land without any tests — we should add a couple. Most valuable would be one asserting the encoded `move` args (that'd catch the records issue above), plus the button-hidden-for-personal-ENS case. `userDialog.test.tsx` is a good template.

src/modules/application/dialogs/aragonProfileRenameTransactionDialog/aragonProfileRenameTransactionDialog.tsx
```diff
@@ -65,11 +84,11 @@ export const AragonProfileRenameTransactionDialog: React.FC<
                 data: encodeFunctionData({
                     abi: memberRegistryAbi,
                     functionName: 'move',
-                    args: [subdomain],
+                    args: [subdomain, records],
```
thekidnamedkd, line 87: `move` is overloaded in the ABI now (`move(string)` + `move(string, Records)`) — viem picks the right one by arg count so this works, just flagging the single-arg version has no callers left. Intentional to keep it around?

src/app/api/subdomain/member-profile/[name]/records/route.ts
```diff
@@ -0,0 +1,38 @@
...
+interface IRequestParams {
+    /**
+     * Aragon profile name, e.g. `test.aragon.eth`
+     */
+    name: string;
+}
+
+interface IRequestOptions {
+    /**
+     * Parameters of the RPC request call.
```
Comment (thekidnamedkd) [inline src/app/api/subdomain/member-profile/[name]/records/route.ts]: Looks like a copy-paste from the RPC routes — this one's Envio GraphQL, not RPC.

```suggestion
     * Parameters of the request call.
```

src/modules/application/api/memberProfileService/memberProfileServiceKeys.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export enum MemberProfileServiceKey {
+    TEXT_RECORDS = 'TEXT_RECORDS',
+}
+
+export const memberProfileServiceKeys = {
+    textRecords: (params: IGetEnsTextRecordsParams) => [
+        MemberProfileServiceKey.TEXT_RECORDS,
+        params,
+    ],
```
thekidnamedkd, line 11: This key skips `apiVersionUtils.getApiVersion()` that our other service keys include — fine if this endpoint is intentionally version-stable, otherwise we should align it so cache busts on version bumps.

src/modules/application/dialogs/aragonProfileRenameDialog/aragonProfileRenameDialog.tsx
thekidnamedkd: One more edge case here: the SDK returns `[]` both when a profile genuinely has no text records and when the name/resolver isn’t found. That means indexer lag or a lookup miss would still enable submit and we could pass an empty `textRecords` array into `move`.

Could we make the records response distinguish “found with zero records” from “not found”, and only enable the rename once we know we loaded the current profile successfully?

src/modules/application/api/memberProfileService/domain/IMemberProfileTextRecord.ts
```diff
@@ -0,0 +1,4 @@
+export interface IMemberProfileTextRecord {
```
asciiman: We should use the DTO from aragon-domain instead. I just noticed that it's not exported though. Let me make a quick PR to export it. Then we can use it.

asciiman: https://github.com/aragon/aragon-domain/pull/16

asciiman: Merged ✅

src/modules/application/api/memberProfileService/queries/useMemberProfileTextRecords/useMemberProfileTextRecords.server.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @example
+ * await queryClient.prefetchQuery(
+ *     memberProfileTextRecordsOptionsServer({ urlParams: { name } }),
+ *  );
+ */
+export const memberProfileTextRecordsOptionsServer = (
+    params: IGetEnsTextRecordsParams,
+    options?: QueryOptions<IMemberProfileTextRecord[]>,
+): SharedQueryOptions<IMemberProfileTextRecord[]> => ({
```
asciiman:
```suggestion
    options?: QueryOptions<MemberProfileTextRecordDTO[]>,
): SharedQueryOptions<MemberProfileTextRecordDTO[]> => ({
```

src/modules/application/api/memberProfileService/memberProfileService.client.ts
```diff
@@ -0,0 +1,20 @@
...
+        const result = await this.request<IMemberProfileTextRecord[]>(
```
asciiman:
```suggestion
        const result = await this.request<MemberProfileTextRecordDTO[]>(
```

src/app/api/domain/member-profile/[subdomain]/records/route.ts
Comment (asciiman) [inline src/app/api/domain/member-profile/[subdomain]/records/route.ts:1]: src/app/api/domain/member-profile/[name]/records/route.ts

src/modules/application/api/memberProfileService/queries/useMemberProfileTextRecords/useMemberProfileTextRecords.ts
```diff
@@ -0,0 +1,20 @@
...
+    options?: QueryOptions<IMemberProfileTextRecord[]>,
+): SharedQueryOptions<IMemberProfileTextRecord[]> => ({
+    queryKey: memberProfileServiceKeys.textRecords(params),
+    queryFn: () => memberProfileServiceClient.getEnsTextRecords(params),
+    ...options,
+});
+
+export const useMemberProfileTextRecords = (
+    params: IGetEnsTextRecordsParams,
+    options?: QueryOptions<IMemberProfileTextRecord[]>,
```
asciiman:
```suggestion
    options?: QueryOptions<MemberProfileTextRecordDTO[]>,
```

src/modules/application/api/memberProfileService/memberProfileService.api.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export interface IGetEnsTextRecordsUrlParams {
+    /**
+     * Aragon profile name, e.g. "test.aragon.eth".
+     */
+    name: string;
+}
+
+export interface IGetEnsTextRecordsParams
+    extends IRequestUrlParams<IGetEnsTextRecordsUrlParams> {}
```
asciiman:
```suggestion
export interface IGetEnsTextRecordsParams
    extends IRequestUrlParams<GetMemberProfileTextRecordsRequestDTO> {}
```

asciiman: Use DTO from aragon-domain

src/modules/application/api/memberProfileService/memberProfileService.server.ts
```diff
@@ -0,0 +1,25 @@
+import 'server-only';
+import { aragonSubdomainServiceBackend } from '@/shared/api/aragonSubdomainService/aragonSubdomainService.backend';
+import type { IGetEnsTextRecordsParams } from './memberProfileService.api';
+
+class MemberProfileServiceServer {
+    getEnsTextRecords = async (params: IGetEnsTextRecordsParams) => {
+        const result =
+            await aragonSubdomainServiceBackend.getMemberProfileTextRecords({
+                subdomain: params.urlParams.name,
+            });
```
asciiman:
```suggestion
    getEnsTextRecords = async ({ urlParams }: IGetEnsTextRecordsParams) => {
        const result =
            await aragonSubdomainServiceBackend.getMemberProfileTextRecords(
                urlParams,
            );
```

asciiman: Can be done like this once the aragon-domain is used inside IGetEnsTextRecordsParams

src/modules/application/dialogs/aragonProfileRenameTransactionDialog/aragonProfileRenameTransactionDialog.tsx
tyhonchik: Intentional — the ABI just mirrors the deployed contract, so we keep both overloads even though only move(string, Records) has callers.

src/shared/api/aragonSubdomainService/aragonSubdomainService.backend.ts
tyhonchik: apiToken is optional on EnvioClient, so no null check needed. The controller now lives in a server-side singleton that initialises lazily, so a missing endpoint surfaces as a clean 500 inside the route instead of an import-time crash.

src/modules/application/dialogs/aragonProfileRenameDialog/aragonProfileRenameDialog.tsx
tyhonchik: Fixed.

We now read the name's current addr and contenthash live on-chain (via the ENS module) and move them through unchanged. Text records still come from aragon-domain since only the indexer can enumerate them. I checked the indexer and aragon-domain: neither stores addr/contenthash (and an indexed value would lag, which is risky for a write), so reading them on-chain is the correct source.

The resolver lookup also gives a real "name found" signal, so submit is only enabled once the current profile actually loaded — no more empty-vs-not-found ambiguity. Added a test asserting the encoded move args.

src/modules/application/api/memberProfileService/memberProfileServiceKeys.ts
tyhonchik: This endpoint is version-stable (Envio/domain, not the versioned Aragon backend), so the key intentionally skips getApiVersion().

src/app/api/domain/member-profile/[subdomain]/records/route.ts
Comment (tyhonchik) [reply src/app/api/domain/member-profile/[subdomain]/records/route.ts:1]: Renamed the dir to `/api/domain` 👍. I kept the dynamic segment as `[subdomain]` instead of `[name]` so the route's `urlParams` match `GetMemberProfileTextRecordsRequestDTO` exactly and the server can pass them straight through to `getMemberProfileTextRecords` (your other suggestion) with no mapping step. Happy to switch back to `[name]` if you prefer that naming.

src/shared/api/aragonSubdomainService/aragonSubdomainService.backend.ts
tyhonchik: Confirmed — both `NEXT_SECRET_ENVIO_GRAPHQL_ENDPOINT` and `NEXT_SECRET_ENVIO_API_TOKEN` are set across all preview/prod envs. On the code side it's also handled now: the controller initialises lazily inside a server-side singleton, so a missing endpoint surfaces as a clean 500 in the route instead of an import-time crash. `apiToken` stays optional, so no null check is needed.

src/modules/application/api/memberProfileService/memberProfileService.client.ts
tyhonchik: Added tests: `aragonProfileRenameTransactionDialog.test.tsx` asserts the encoded `move` args carry the full record set (text + addr + contenthash — guards the data-loss fix), `aragonProfileRenameDialog.test.tsx` covers the loading / error / not-found gating, and `aragonProfileDialog.test.tsx` covers the rename button being hidden for non-Aragon ENS names.

src/app/api/subdomain/member-profile/[name]/records/route.ts
Comment (tyhonchik) [reply src/app/api/subdomain/member-profile/[name]/records/route.ts]: Done — updated to "Parameters of the request call."

src/modules/application/api/memberProfileService/domain/IMemberProfileTextRecord.ts
tyhonchik: Done — dropped the local `IMemberProfileTextRecord` and now use `MemberProfileTextRecordDTO` from `@aragon/aragon-domain` everywhere (client, server, queries).

thekidnamedkd (approved):
lgtm 🔥

PR #1189, opened 2026-06-23
feat(APP-666): Implement pending transaction management for transaction dialog

harryburger:
# Description

Reduces the chance of accidentally sending a **duplicate onchain transaction** (e.g. a duplicate proposal) when the transaction dialog is closed mid-flow and then reopened. Implements **APP-666**.

### Problem

The transaction dialog can be dismissed with `Esc` while a request is awaiting the user's signature in their wallet. Because the dialog itself owned the send, closing it lost all awareness that a request had already been handed to the wallet — reopening would re-run the flow and send a **second** transaction.

### Approach — own the send outside the dialog, split into single-responsibility layers

The in-flight wallet send is owned by a module-level singleton (`pendingTransactionManager`) rather than the dialog, keyed by a stable per-action identity (`intentId`). Because the singleton — not the React component — awaits wagmi's core `sendTransaction` promise, the sign/reject outcome survives the dialog unmounting.

- **`pendingTransactionManager`** — owns sends + state only (no React, no telemetry). Transitions `PENDING → SUBMITTED / FAILED`, mirrors `{status, hash}` to `sessionStorage`, and exposes a change stream. A per-intent **attempt token** makes a stale (out-of-order) resolution a no-op, and the last request is retained so a resumed dialog can re-send.
- **`transactionLoggingSubscriber`** — an independent subscriber, wired once at app init, that reports each `FAILED` send exactly once — **even when the dialog is closed** — and never re-logs on step navigation. The manager imports no telemetry.
- **`useManagedTransaction(intentId)`** — the single hook binding UI to the manager: store subscription, hash latch, receipt wait, status→display mapping, a **synchronously-computed resume target** (no effect-ordering race), and `send` / `resend`.
- **`TransactionDialog`** — now a thin renderer: reads the hook, maps to steps, and dispatches `send` / `resend`. No manager import, no `useSyncExternalStore`, no resume bookkeeping.

On reopen the dialog resumes where the prior attempt left off:
- `SUBMITTED` (signed) → resume at **Confirm** on the stored hash (survives a full page reload too)
- live `PENDING` (awaiting signature) → resume at **Sign**, still attached to the live request; after a rejection, **Resend** re-sends the stored request
- a `PENDING` interrupted by a reload (no live promise → unknown outcome) → start fresh, never auto-send

`intentId` is a **required** prop, derived at each call site from the action's defining inputs (`buildIntentId`) so it is the same across re-opens and available before `prepare` runs — which is what makes resume deterministic.

### Acceptance criteria

- **AC1** — an unresolved request already handed to the wallet is not easy to accidentally re-send (the dialog resumes rather than re-sends).
- **AC2** — closing and reopening does not erase the app's awareness that the action was handed to the wallet (it shows the waiting/confirm state).
- **AC3** — the user can still retry if the original request was rejected, dismissed, or lost.

### Review

Folds in the Copilot review (out-of-order resolution → per-intent attempt token; `FAILED` re-logging on step nav → moved to the always-alive subscriber; stale doc / misleading error message → corrected) and the architectural review (four single-responsibility layers; required `intentId`; explicit `buildIntentId` field whitelists; latch reset so a retry can't advance on a stale hash; resend surfaces a failure instead of a silent no-op).

### Tests

Unit tests cover the manager (lifecycle, attempt token, hydrate guard, request retention), the `useManagedTransaction` hook (resume branches, latch + latch-reset, status mapping, send/resend), the logging subscriber (log-once), and the dialog (send/resume/resend + surfaced errors).

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/src/shared/components/transactionDialog/transactionDialog.tsx
+++ b/src/shared/components/transactionDialog/transactionDialog.tsx
@@ -195,9 +195,12 @@ export const TransactionDialog = <TCustomStepId extends string>(
             ],
         );
 
-    const approveStepStatus = isCrossNetworkTransaction
-        ? switchChainStatus
-        : approveState;
+    // Let a managed send win over the chain-switch state: a resumed pending/submitted action must not
+    // look idle (which would invite a duplicate re-send) just because the wallet is on another chain.
+    const approveStepStatus =
+        isCrossNetworkTransaction && approveState === 'idle'
+            ? switchChainStatus
+            : approveState;
     const indexingStepStatus = transactionStatus?.isProcessed
         ? 'success'
         : isIndexing
--- a/src/shared/components/transactionDialog/transactionDialog.tsx
+++ b/src/shared/components/transactionDialog/transactionDialog.tsx
@@ -7,6 +7,7 @@ import { useTransactionStatus } from '@/shared/api/transactionService';
 import { useDialogContext } from '@/shared/components/dialogProvider';
 import { useDaoChain } from '@/shared/hooks/useDaoChain';
 import { useNetworkSwitch } from '@/shared/hooks/useNetworkSwitch';
+import { buildIntentId } from '@/shared/utils/pendingTransactionManager';
 import { NetworkSwitchAlert } from '../networkSwitchAlert';
 import {
     type ITransactionStatusStepMetaAddon,
@@ -30,7 +31,7 @@ export const TransactionDialog = <TCustomStepId extends string>(
     const {
         title,
         description,
-        intentId,
+        intentId: intentIdProp,
         customSteps,
         transactionInfo,
         stepper,
@@ -91,7 +92,23 @@ export const TransactionDialog = <TCustomStepId extends string>(
         data: transaction,
     } = useMutation({ mutationFn: prepareTransaction, onSuccess: nextStep });
 
-    // The wallet send + its resume state live in the manager, surfaced through this hook.
+    // Resume identity: explicit prop when the calldata is non-deterministic (e.g. proposals),
+    // otherwise derived from the prepared transaction, which is stable for the same action.
+    const intentId = useMemo(
+        () =>
+            intentIdProp ??
+            (transaction != null && address != null
+                ? buildIntentId({
+                      from: address,
+                      chainId: requiredChainId,
+                      to: transaction.to,
+                      data: transaction.data,
+                      value: transaction.value,
+                  })
+                : undefined),
+        [intentIdProp, transaction, address, requiredChainId],
+    );
+
     const { approveState, hash, resumeTarget, receipt, send, resend } =
         useManagedTransaction(intentId);
     const {
@@ -116,24 +133,41 @@ export const TransactionDialog = <TCustomStepId extends string>(
     });
 
     const handleSendTransaction = useCallback(() => {
-        // A resumed dialog skipped prepare, so it has no freshly-built transaction — re-send the
-        // request the manager kept from the original send. Surface a failure if there is none (e.g.
-        // after a reload, where the in-memory request is gone) rather than no-op silently.
+        const onError = handleTransactionError(TransactionDialogStep.APPROVE);
+
+        // No prepared transaction = a resumed dialog that skipped prepare: re-send the stored
+        // request, or surface a failure if there is none (e.g. lost after a reload).
         if (transaction == null) {
             if (!resend()) {
-                handleTransactionError(TransactionDialogStep.APPROVE)(
+                onError(
                     new Error(
-                        'TransactionDialog: no request available to re-send.',
+                        'TransactionDialog: no prepared transaction and nothing to re-send.',
                     ),
                 );
             }
             return;
         }
 
-        // Pin to the required chain so wagmi rejects (rather than silently signing) if the wallet is
-        // still on the wrong one.
+        // With a prepared transaction the id is always derivable; a missing one means no connection.
+        if (intentId == null) {
+            onError(
+                new Error(
+                    'TransactionDialog: cannot identify the action to send (wallet not connected).',
+                ),
+            );
+            return;
+        }
+
+        // Pin to the required chain so wagmi rejects instead of silently signing on the wrong one.
         send({ ...transaction, chainId: requiredChainId });
-    }, [transaction, requiredChainId, send, resend, handleTransactionError]);
+    }, [
+        transaction,
+        intentId,
+        requiredChainId,
+        send,
+        resend,
+        handleTransactionError,
+    ]);
 
     const handleRetryTransaction = useCallback(() => {
         updateActiveStep(TransactionDialogStep.APPROVE);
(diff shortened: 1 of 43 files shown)


tyhonchik (changes requested):
Really like this direction — owning the in-flight send outside the dialog so it survives close/reopen is exactly right, and the Copilot notes (out-of-order resolution, attempt tracking cleanup, misleading error message, stale `intentId` doc) are all valid and worth folding in. Thanks for pushing this.

Before we lock the shape, here's how I think it'd be cleaner / more reliable architecturally — and it happens to resolve those bot comments as a side effect. The core idea: split the concern into three layers with single responsibilities, so the UI and the manager stay simple and decoupled.

```mermaid
flowchart LR
    UI["TransactionDialog (dumb UI)"] -->|"intentId + actions"| Hook["useManagedTransaction (React binding)"]
    Hook -->|"subscribe / send / get"| Mgr["pendingTransactionManager (state + sends)"]
    Logger["txLoggingSubscriber (telemetry)"] -->|"subscribe"| Mgr
    Logger --> Mon["monitoringUtils (Sentry boundary)"]
```

### 1. Manager — owns sends + state only
No React, no Sentry. It launches the send, transitions `PENDING → SUBMITTED / FAILED`, persists, and exposes `subscribe`. An internal per-intent attempt token makes stale resolutions no-ops (resolves the out-of-order bug + attempt cleanup).

```ts
class PendingTransactionManager {
    send(intentId, request): void   // attempt-token inside; old promise can't overwrite newer state
    get(intentId): IPendingTransactionState | undefined
    isInterrupted(intentId): boolean
    clear(intentId): void
    subscribe(listener): () => void
}
```

### 2. Logging — a separate subscriber
The manager just emits changes; telemetry is an independent module wired once next to the singleton. The manager never imports `monitoringUtils`, and because the subscriber is always alive it logs a `FAILED` outcome **even when the dialog is closed** (today that case is lost) — without re-logging on step navigation.

```ts
// txLoggingSubscriber.ts — wired once at module init
pendingTransactionManager.subscribe(() => {
    // log each FAILED transition exactly once; beforeSend tags wallet rejections noise_class=expected
});
```

### 3. `useManagedTransaction(intentId)` — the single UI↔manager binding
All derivation lives here in one place: `useSyncExternalStore` + a `ts-pattern` `match` from status to display phase, the synchronously-computed resume target, and the `send`/`retry` actions. The dialog depends on this hook's interface, not on the singleton.

```ts
const { phase, hash, resumeTarget, send, retry } = useManagedTransaction(intentId);
// phase = match(status).with(SUBMITTED, () => 'success')... .exhaustive()
// resumeTarget via useState(() => ...) — computed once, no effect-ordering race
```

### 4. Dialog — dumb rendering
Reads `phase`/`resumeTarget` from the hook, maps to steps, calls `send`/`retry`. No direct manager import, no `useSyncExternalStore`, no `latchedHash`, no three ordered effects to sequence resume.

### Why I'd go this way

Each layer does one thing: manager — state, subscriber — logging, hook — UI binding, dialog — render. Less overlap, fewer bugs from effect ordering.

Logging in a subscriber that’s always alive means a wallet rejection while the dialog is closed still gets reported. No duplicate logs on step changes.

Status → UI mapping lives in one hook, not spread across ternaries and effects. Easier to read and change.

Later: new reaction on failure = new subscriber; new status = one `match` branch. Required `intentId` makes resume on reopen reliable — no waiting for `prepare` to finish first.

### Two small follow-ups
- Make `intentId` a **required prop** (as `publishProposalDialog` already passes). It's currently auto-derived from `transaction`, which only exists after `prepare`, so there's no id to resume a `PENDING` from on reopen — a passed-in stable id makes resume deterministic and drops the auto-derive branch.
- `buildIntentId` for proposals hashes the whole object — hash an explicit field whitelist so a volatile field can't make the id drift; and add a small enum guard in `hydrate` against stale storage.

tyhonchik (commented):
Did small additional changes

tyhonchik (dismissed):

src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -166,7 +197,7 @@ export const TransactionDialog = <TCustomStepId extends string>(
 
     const approveStepStatus = isCrossNetworkTransaction
         ? switchChainStatus
-        : approveTransactionStatus;
+        : approveState;
```
thekidnamedkd: I think we should let the managed pending state win over the network switch state here, because if someone reopens a pending transaction after switching chains this can look idle and let them resend the stored request.

```suggestion
const approveStepStatus =
    isCrossNetworkTransaction && approveState === 'idle'
        ? switchChainStatus
        : approveState;
```

src/shared/components/transactionDialog/useManagedTransaction.ts
```diff
@@ -0,0 +1,121 @@
...
+            setResumeTarget(TransactionDialogStep.CONFIRM);
+        } else if (
+            status === PendingTransactionStatus.PENDING &&
+            !pendingTransactionManager.isInterrupted(intentId)
+        ) {
+            setResumeTarget(TransactionDialogStep.APPROVE);
+        } else if (status != null) {
+            pendingTransactionManager.clear(intentId);
+        }
+    }, [intentId]);
```
thekidnamedkd, line 83: I think we should reset the latched hash and resume target whenever the intent id changes, otherwise a reused dialog instance can carry the previous action’s hash or resume step into a fresh action.

```suggestion
    useEffect(() => {
        setLatchedHash(undefined);
        setResumeTarget(undefined);

        if (intentId == null) {
            return;
        }

        const status = pendingTransactionManager.get(intentId)?.status;
        if (status === PendingTransactionStatus.SUBMITTED) {
            setResumeTarget(TransactionDialogStep.CONFIRM);
        } else if (
            status === PendingTransactionStatus.PENDING &&
            !pendingTransactionManager.isInterrupted(intentId)
        ) {
            setResumeTarget(TransactionDialogStep.APPROVE);
        } else if (status != null) {
            pendingTransactionManager.clear(intentId);
        }
    }, [intentId]);
```

src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -301,6 +361,13 @@ export const TransactionDialog = <TCustomStepId extends string>(
         }
     }, [waitTxStatus, nextStep, txReceipt]);
 
+    // Advance to confirm once the wallet has signed (the hash appears). Failures go to the subscriber.
+    useEffect(() => {
+        if (hash != null && activeStep === TransactionDialogStep.APPROVE) {
+            nextStep();
+        }
+    }, [hash, activeStep, nextStep]);
```
thekidnamedkd, line 369: I am little nervous about this many side effects. Thoughts?

src/shared/components/transactionDialog/transactionLoggingSubscriber.ts
```diff
@@ -0,0 +1,41 @@
...
+
+let initialized = false;
+
+export const initTransactionLogging = (): void => {
+    if (initialized) {
+        return;
+    }
+    initialized = true;
+    pendingTransactionManager.subscribe(createTransactionLogger());
+};
```
thekidnamedkd, line 41: is there any teardown here? necessary?

thekidnamedkd (commented):
looking good, a couple comments

src/shared/components/transactionDialog/transactionDialog.tsx
tyhonchik: each effect syncs a single concern into `useStepper` API and has its own dependency array. merging them would broaden deps and re-trigger step jumps mid-flow, so I've grouped and labeled them instead

src/shared/components/transactionDialog/transactionLoggingSubscriber.ts
tyhonchik: this is a process-lifetime singleton wired once at init, and loggedIntents is pruned on every non-FAILED transition, so nothing accumulates. Tearing it down would drop failure logging while the dialog is closed, which is the case we want to keep covered.

thekidnamedkd (approved):
lgtm 🔥

PR #1191, opened 2026-06-23
feat(APP-962): route Sentry signal, suppress only true noise

tyhonchik:
## Context

Sentry for `app-next` was noisy: the top issues were bot/scanner traffic and expected user behaviour, real bugs were buried, every issue showed `Users: 0` (no `setUser`), and there was no written strategy for what to report. This makes Sentry conscious: it routes signal, suppresses only true noise, and keeps everything else searchable for investigation.

## What changed

**Monitoring (`monitoringUtils.beforeSend` — applies to client/server/edge)**
- **Drops only zero-signal third-party junk**: browser-extension internals, mail crawlers, and in-app-browser/WebView scripts that `JSON.stringify` DOM nodes (matched narrowly via `__reactFiber`, so a genuine cyclic bug in our own data still surfaces).
- **Routes, doesn't hide.** Everything that could aid investigation is kept and tagged via `noise_class`:
  - `expected` — user/wallet behaviour (rejection, insufficient balance, not connected, WalletConnect stale session, EIP-1193 codes) **and** non-actionable external failures (ENS off-chain/CCIP gateway). Demoted to `level: info` — searchable by `user.id:0x…` but never alerts.
  - `infra` — backend/RPC/proxy failures (502, RPC error, HTML-instead-of-JSON). Visible, for devops/backend.
  - `security-probe` — scanner/injection URLs (e.g. SSTI payloads). Kept for security review.
- **`setUser`** via new `SentryUserSync` (`monitoringUtils.identifyUser`) — fixes `Users: 0`, enables `user.id` lookup + Session Replay.

**Expected not-found (stale links to DAOs whose plugins were removed)**
- `AragonBackendServiceError.isExpectedNotFoundError` (`notFound` + `pluginNotFound`); `pageError` and the metadata utils render a clean not-found page and stop reporting these.

**Crash fixes**
- `getNextPageParams`: a null infinite-query page no longer throws on `.metadata` (APP-NEXT-217).
- `daoVisibilityUtils.filterHiddenPlugins`: undefined `dao.plugins` no longer throws on `.filter` (APP-NEXT-27Q).

**Convention**
- New `error-and-monitoring` rule-skill codifying the taxonomy so new features wire logging correctly.

## Sentry triage (done via MCP)
~24 issues archived with rationale (noise / expected / probe / external / stale-link / WebView-cyclic), real bugs kept/assigned, backend cluster annotated for hand-off, fixed crashes resolved.

## Follow-ups in Sentry UI (cannot be done via code/MCP)
- Saved views: `is:unresolved !noise_class:[expected,infra,security-probe]` (bugs), `noise_class:security-probe`, `noise_class:infra`, `noise_class:expected`.
- Add a `!noise_class:[...]` + `level:error` filter to the issue alert rules so tagged noise never notifies.
- Set issue priorities; route the `infra` + `Bad parameters` (248/24C) + N+1 (APP-388) cluster to backend.
- Tags populate on **new** events post-deploy.

## Verification
`pnpm type-check` ✓ · tests ✓ (203 app + 30 guardrails) · `ultracite` ✓.

### DIFF
--- a/src/modules/application/utils/applicationMetadataUtils/applicationMetadataUtils.ts
+++ b/src/modules/application/utils/applicationMetadataUtils/applicationMetadataUtils.ts
@@ -29,7 +29,16 @@ class ApplicationMetadataUtils {
                 // fixing. Tag both via `noise_class` so they route out of the default alert
                 // stream (internal-broken-link → triage, security-probe → security review).
                 const referer = (await headers()).get('referer') ?? '';
-                const isInternalLink = referer.includes('aragon.org');
+                let isInternalLink = false;
+
+                try {
+                    const refererHost = new URL(referer).hostname.toLowerCase();
+                    isInternalLink =
+                        refererHost === 'aragon.org' ||
+                        refererHost.endsWith('.aragon.org');
+                } catch {
+                    isInternalLink = false;
+                }
 
                 monitoringUtils.logMessage('Invalid DAO URL', {
                     level: isInternalLink ? 'warning' : 'info',
--- a/src/modules/application/utils/applicationMetadataUtils/applicationMetadataUtils.ts
+++ b/src/modules/application/utils/applicationMetadataUtils/applicationMetadataUtils.ts
@@ -1,4 +1,5 @@
 import type { Metadata } from 'next';
+import { headers } from 'next/headers';
 import { AragonBackendServiceError } from '@/shared/api/aragonBackendService';
 import { daoService } from '@/shared/api/daoService';
 import type { IDaoPageParams } from '@/shared/types';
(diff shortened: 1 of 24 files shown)


src/shared/utils/monitoringUtils/monitoringUtils.ts
```diff
@@ -25,23 +32,143 @@ export interface ILogMessageParams {
...
+     * Backend/RPC/proxy failures. Real signal, but the source is outside our codebase —
+     * tagged `infra` so devops/backend can pick them up, kept visible (never dropped).
+     */
+    private infraPatterns = [
+        "Unexpected token '<'", // Backend/proxy returned HTML instead of JSON
+        '<html>',
+        'Error parsing response',
+        'RPC endpoint returned error status',
+        'HTTP request failed',
+        'fetch failed',
```
thekidnamedkd, line 122: Since we match these as substrings across the whole concatenated message, we could accidentally tag a real bug as `infra` (or `expected`, for the phrases up in `expectedErrorPatterns` like `must be connected` / `exceeds the balance`) if its message just happens to contain one of these strings. It might be worth anchoring the broadest phrases like `must be connected`) to an error type/code down the road rather than free text. 🙂

src/modules/application/utils/applicationMetadataUtils/applicationMetadataUtils.ts
```diff
@@ -23,10 +24,31 @@ class ApplicationMetadataUtils {
...
+                    isInternalLink = false;
+                }
+
                 monitoringUtils.logMessage('Invalid DAO URL', {
+                    level: isInternalLink ? 'warning' : 'info',
                     context: {
                         network: daoPageParams.network,
                         addressOrEns: daoPageParams.addressOrEns,
+                        referer,
+                        noise_class: isInternalLink
```
thekidnamedkd: here we're setting `noise_class` as a free-string `logMessage` context tag, which skips `beforeSend` and uses values that aren't in our `SentryNoiseClass` union (`internal-broken-link` isn't a member). So we're now tagging noise two different ways—typed + central in `beforeSend`, and ad-hoc here—which means this path can drift from the saved views / alert filters we're setting up elsewhere.

thekidnamedkd (approved):
fuego 💧

PR #1198, opened 2026-06-30
feat(APP-954): Add permissions overview subpage to DAO settings

thekidnamedkd:
# Description

Adds the permissions overview subpage (**list view**) to DAO settings. APP-954, child of APP-321.

- New route `/dao/[network]/[addressOrEns]/settings/permissions` + a "Permissions" entry link on the settings page.
- Linked-account tabs; permission rows (Who / Where / Permission / Condition) as an accordion; expand → two definition lists (**Details** | **Condition detail**); skeleton + empty states.
- Condition slot system: standalone `initConditionRegistry` (no plugin-index edits), routed via `PluginSingleComponent` with a `NoConditionSlot` fallback; VotingPower + ExecuteSelector condition components.
- 65-entry permission-id → name dictionary (monospace); Who/Where entity resolution with plugin type tags (MULTISIG / SPP).
- Preview-mocked via the `useMocks` seam (ahead of backend **APP-953**). Verified against Figma (list + variants) at desktop + mobile.

**Out of scope (split out):** Graph view → APP-967 (toggle ships present-but-disabled); polished entry points (metadata accordion / dashboard card) → APP-955; live data + generic condition resolution → APP-953.

## Type of Change

## Developer Checklist:


harryburger (dismissed):

tyhonchik (changes requested):
Some notes before this can merge. Mix of blocking and general.

## 1. e2e is actually red (8 tests failed)
The e2e check shows green, but that's only because e2e doesn't block the PR. The run is `71 passed, 8 failed, 1 skipped`. All 8 failures are the same DAO settings smoke test across different DAOs:

`e2e/tests/smoke/daoPage/daoSettings.spec.ts:21 › Governance process detail renders` — the "Governance process" heading never appears.

This page is one this PR touches, so it looks like a regression from here. It needs to be really green before merge.

## 2. Global mock is too broad — likely the cause of the e2e failures
`src/backendApiMocks.ts` now registers `permissionsMocks` app-wide, and the match is a regex on `/permissions/` with `type: 'replace'`. When `useMocks` is on, this can intercept other requests across the app, not just this page. Let's scope the mock to this feature only and keep it behind the same flag as the page, so it can't affect other pages.

## 3. The page has no real backend yet — gate it
Real data is APP-953, which isn't here. Right now the route and the entry links are live, so on production this hits a non-existent endpoint and renders broken. Let's put the route and the entry points behind a feature flag and enable it only when the backend is ready. Merging the code is fine; showing a half-page with mock data to users is not.

## 4. Link opens a new tab
The Permissions link opens in a new tab. It's an internal link and should stay in the same tab.

## 5. Design doesn't match Figma
[Figma](https://www.figma.com/design/mnH992k1U8hHBzTHFC2W3U/Aragon-App?node-id=21334-247840&m=dev)

- View toggle / tabs: position and styles differ from the design.
- Who/Where contracts (DAO + plugins) aren't shown with proper names, and the pills/tags are missing.
- Layout/spacing in general.
- Entrypoint link style.

Needs a pass to match Figma, and so on.

## 6. Helpers duplicated (probably already in utils somewhere?)
Several small helpers are defined locally that we likely already have shared:

- `isNonEmptyString` is copied in two slot files.
- `permissionNameUtils` re-implements `truncateHash`, while `addressUtils.truncateHash` is used right next to it.
- `toPascalCase`, etc.

Let's reuse the shared utils instead of local copies.

## 7. Permission ids/names are scattered
The permission hashes are inline in the mocks and the name list is a big inline array. Let's keep the permission dictionary in one place so it's easy to find and extend.

## 8. The ADR file
We don't use ADRs in this repo. Adding one inside a feature PR introduces a new convention on its own. Let's drop it here, or discuss separately if we actually want ADRs

---

Overall I'm suggesting: flag it off + hide the links, fix the e2e, align with Figma, etc. then it's good to go.

tyhonchik (approved):
Nice, all points are addressed
Thanks for the quick turnaround!
🙌 LGTM ✅

PR #1200, opened 2026-06-30
feat(APP-816): validate Aragon profile fields and show all socials

tyhonchik:
# Description

Resolves **APP-816** — Aragon profile fields are not updated correctly / no form validation.

Aragon member profiles write contact/social fields to ENS text records (ENSIP-5 / ENSIP-18) and read them back for display. This PR makes that round-trip consistent and safe:

### Edit profile validation
Added react-hook-form validation rules for **all** supported profile/contact fields, surfacing inline errors (via the existing `useFormField` alert plumbing) and blocking submission of malformed values:

- **Email** — validated as an email address
- **Website** — validated as an http/https URL (reuses `sanitizeExternalHttpUrl`)
- **GitHub / X / Telegram / Discord** — validated as service usernames per ENS profile-text-record expectations

Empty values stay valid (fields are optional/removable). Twitter/Telegram handles are normalized (leading `@` stripped) before being written to ENS.

### Profile display
- The member profile page now displays **Email, Telegram and Discord** when present, alongside the existing **Website / X / GitHub** (Email → `mailto:`, Telegram → `t.me`, Discord → plain text since usernames have no canonical URL).
- The **Links** card is now only rendered when at least one link resolves — fixing the bug where it showed as an **empty card**. A shared `buildMemberLinks` helper is the single source of truth for both the card and the parent's render gate.
- All values are fetched from ENS the same way as the existing fields (`useEnsProfileRecords` already fetched all six keys).

## Type of Change

## Developer Checklist:

**Verification done:** `pnpm jest` (memberLinksCard + aragonProfileValidation → 60 passing, incl. new tests), `pnpm type-check` clean, `pnpm lint:check` clean. Manual smoke test in a preview still pending.

### DIFF
--- a/src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
+++ b/src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
@@ -35,7 +35,13 @@ export const AragonProfileSocialFieldRow: React.FC<
 
     const socialField = useFormField<IAragonProfileDialogFormData, SocialKey>(
         fieldName,
-        { trimOnBlur: true, rules: socialFieldRules[fieldName] },
+        // `sanitizeOnBlur` is required for `trimOnBlur` to run, so pasted values like
+        // `alice@example.com ` or `@aragon ` are normalized before validation.
+        {
+            trimOnBlur: true,
+            sanitizeOnBlur: true,
+            rules: socialFieldRules[fieldName],
+        },
     );
 
     return (
@@ -45,7 +51,7 @@ export const AragonProfileSocialFieldRow: React.FC<
                 iconLeft={socialKeyToIconType[fieldName]}
                 id={`aragon-profile-social-${fieldName}`}
                 placeholder={t(
-                    `app.application.aragonProfileDialog.fields.socials.${fieldName}`,
+                    `app.application.aragonProfileDialog.fields.socials.placeholders.${fieldName}`,
                 )}
                 {...socialField}
             />
--- a/src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
+++ b/src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
@@ -7,6 +7,7 @@ import type {
     IAragonProfileDialogFormData,
     SocialKey,
 } from './aragonProfileDialog';
+import { socialFieldRules } from './aragonProfileValidation';
 
 export interface IAragonProfileSocialFieldRowProps {
     /** Form field name. */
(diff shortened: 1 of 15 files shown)


src/modules/governance/components/memberLinksCard/memberLinksCard.tsx
```diff
@@ -43,21 +51,44 @@ export function buildGithubUrl(handle: string): string | null {
...
- * Returns `null` when no links are available so the parent can skip rendering the card.
+ * Builds a safe Telegram profile URL from a handle.
+ * Strips a leading `@` if present and encodes the handle to prevent URL injection.
  */
-export const MemberLinksCard: React.FC<IMemberLinksCardProps> = (props) => {
-    const { url, twitter, github } = props;
-    const { t } = useTranslations();
+export function buildTelegramUrl(handle: string): string | null {
+    const cleaned = handle.startsWith('@') ? handle.slice(1) : handle;
+    if (!SAFE_HANDLE_RE.test(cleaned)) {
```
thekidnamedkd: What if `buildTelegramUrl` reused the same Telegram constraints as the form validation? The profile page can render ENS records that were created before this change or written outside Aragon, and the generic handle regex still turns values like `abc`, `foo.bar`, or `foo-bar` into dead `t.me` links even though the save path rejects them. Matching the `^\w{5,32}$` check after stripping `@` would keep display and validation aligned.

src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
```diff
@@ -34,7 +35,7 @@ export const AragonProfileSocialFieldRow: React.FC<
 
     const socialField = useFormField<IAragonProfileDialogFormData, SocialKey>(
         fieldName,
-        { trimOnBlur: true },
+        { trimOnBlur: true, rules: socialFieldRules[fieldName] },
```
thekidnamedkd: What if we enable blur sanitization here so `trimOnBlur` actually runs? `useFormField` currently returns before trimming unless `sanitizeOnBlur` is true, so pasted values like `alice@example.com ` or `@aragon ` still fail these validators. Something like `{ trimOnBlur: true, sanitizeOnBlur: true, rules: socialFieldRules[fieldName] }` should keep the current behavior but normalize the common paste case.

src/modules/governance/components/memberLinksCard/memberLinksCard.tsx
tyhonchik: Good catch. Pulled all handle formats into a shared `socialHandleUtils` so display and validation share one source of truth

src/modules/application/dialogs/aragonProfileDialog/aragonProfileSocialFieldRow.tsx
tyhonchik: Done — added `sanitizeOnBlur: true` so `trimOnBlur` actually runs.

thekidnamedkd (approved):
lgtm 👍

PR #1201, opened 2026-07-01
feat(APP-970): Rename tx dialog Cancel to Close and warn on duplicate proposal creation

tyhonchik:
# Description

Follow-ups from the [#1189](https://github.com/aragon/app/pull/1189) (APP-666) review. Ships the two low-risk items; the dynamic wizard CTA (feedback #2) is deferred to **APP-20** because its main value — a "View proposal" state after confirmation — needs a retained terminal record and a page-persistent widget, which is exactly what APP-20 describes.

### 1/ Rename "Cancel" → "Close"

The transaction dialog now keeps tracking the send in the background after it's dismissed, so the secondary button reading "Cancel" was misleading. Renamed to "Close" (the indexing-timeout "Continue in background" label is unchanged).

### 3/ Warn before creating a duplicate proposal

**Problem:** the proposal `intentId` is derived from the form content. Closing the dialog and then editing the form produces a **new** `intentId`, so the prior in-flight send is no longer resumed and a fresh submit creates a **second proposal**. An unchanged form keeps the same `intentId` and resumes as before.

**Approach:** at submit we recompute the proposal identity and, if another **PENDING/SUBMITTED** proposal creation for the **same DAO + plugin** is still in flight (different `intentId`), we open a warning alert instead of publishing. The safe "Keep waiting" action is primary; "Create anyway" proceeds to publish. We don't clear the old tracking — a SUBMITTED tx is already on-chain and a PENDING wallet request can't be cancelled, so the warning is informed consent.

To support this the pending-transaction manager now stores optional, generic `type`/`scope` metadata (surviving status transitions and reload) and exposes `getActive({ type, scope, excludeIntentId })`. `buildProposalIntentId` / `buildProposalScope` are shared between the dialog and the guard so both compute the identical identity.

### Deferred to APP-20

- Dynamic wizard CTA reflecting live tx status (Sign transaction / View proposal).
- Retained terminal (success) state in the manager + the global, page-persistent transaction widget.

## Type of Change

## Tests

Unit tests cover the new manager query + meta persist/hydrate, meta threading through `useManagedTransaction`, the create-proposal guard branch (warn + proceed), and the alert dialog render/actions. Existing `send` assertions updated for the new `meta` argument.


src/shared/utils/pendingTransactionManager/pendingTransactionManager.api.ts
```diff
@@ -13,6 +13,29 @@ export interface IPendingTransactionState {
...
+export interface IPendingTransactionMeta {
+    type?: string;
+    scope?: string;
+}
+
+// Filter for querying active (PENDING/SUBMITTED) records; every provided field must match.
+export interface IPendingTransactionFilter {
+    type?: string;
+    scope?: string;
+    excludeIntentId?: string;
```
thekidnamedkd: Comments pattern here is neither/or

thekidnamedkd (commented):
One annoying comment about comments. Looks great!

src/shared/utils/pendingTransactionManager/pendingTransactionManager.api.ts
tyhonchik: made it consistent: dropped the per-field JSDoc

thekidnamedkd (dismissed):
lgtm 🍾

thekidnamedkd (approved):
nice 🥇

PR #1203, opened 2026-07-02
fix(APP-970): stop stale duplicate-proposal warnings, resume in-flight transactions

tyhonchik:
Follow-up fix to #1201. While testing proposal creation, the duplicate-proposal warning fired falsely and **never cleared** — most reliably on linked DAOs, where every submit warned. Root cause: the pending-transaction manager accumulated orphaned "active" records that no code path ever cleared.

## Root causes
- **Interrupted PENDING counted as a duplicate.** A PENDING send has no tx hash and its live wallet promise can't survive a reload, so any PENDING restored from `sessionStorage` was a dead ghost — yet the create-page guard's `getActive` counted it as in-flight. It never opened the dialog for that intent, so it never cleared → permanent false warning for the whole `daoId:pluginAddress` scope.
- **Abandoned SUBMITTED never cleared.** The only clear paths were inside `useManagedTransaction`, keyed to the exact `intentId`. Since `intentId` is content-derived, editing the form made a new id while the old SUBMITTED lingered; "Publish again" cleared nothing.
- The warning was generic — it never said *which* transaction it meant.

## Changes
- **Persist only SUBMITTED records.** PENDING is no longer written (it can't be resumed after a reload), which removes the `liveSends`/`isInterrupted` machinery entirely and eliminates the ghost class.
- **Reconcile on load.** Persisted SUBMITTED records are checked against the chain (`getTransactionReceipt`) on hydrate; already-mined ones self-clear, so stale warnings don't survive a reload / cache clear.
- **`clearActive()` on supersede.** Choosing "Publish anyway" clears the conflicting in-flight creation(s) for the scope instead of warning forever.
- **Warning with details + Return.** Lists each in-flight proposal with its status (awaiting signature / sent) and a transaction link, and offers **Return to transaction** to reopen and resume its dialog (session-scoped via a small `proposalResumeRegistry`).
- **Simplification.** Grouped `intentId`/`intentScope`/`intentLabel` into one `intent` prop, dropped the unused `network` meta field, and expressed the resume mapping as an explicit `match`-based state machine with a documented lifecycle.

Works for both regular and linked-account proposals — the dedup scope is `daoId:pluginAddress` in all cases; linked DAOs were only where the ghosts surfaced.

## Testing
- `pnpm type-check` — clean.
- `pnpm jest` — 117 passing across the touched suites; added tests for `clearActive`, hydrate reconciliation, PENDING-not-persisted, label round-trip, and the warning's details/Return.


.changeset/fix-duplicate-proposal-warning.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Fix stale "duplicate proposal" warnings on the create-proposal page: pending wallet requests that were interrupted (page reloaded mid-sign) or already mined are no longer counted as in-flight, and settled transactions are reconciled and cleared on reload. The warning now lists the in-flight proposal(s) with their status and a link to the transaction, and lets you return to the in-flight transaction dialog to resume it or publish anyway to supersede it.
```
thekidnamedkd: Maybe a little verbose for a changelog.

src/modules/governance/utils/proposalResumeRegistry/proposalResumeRegistry.ts
```diff
@@ -0,0 +1,23 @@
...
+class ProposalResumeRegistry {
+    private entries = new Map<string, IPublishProposalDialogParams>();
+
+    set = (intentId: string, params: IPublishProposalDialogParams): void => {
+        this.entries.set(intentId, params);
+    };
+
+    get = (intentId: string): IPublishProposalDialogParams | undefined =>
+        this.entries.get(intentId);
+}
```
thekidnamedkd, line 21: Not exactly a utils, but historically classes get tests.

src/shared/utils/pendingTransactionManager/pendingTransactionManager.ts
```diff
@@ -206,6 +216,28 @@ export class PendingTransactionManager {
...
+                state.hash == null
+            ) {
+                continue;
+            }
+            getTransactionReceipt(wagmiConfig, { hash: state.hash })
+                .then(() => this.clear(id))
+                .catch(() => {
+                    // Not mined yet (or unreadable) — keep it as in-flight.
+                });
+        }
```
thekidnamedkd, line 240: Nice — reconcile-on-hydrate is a great touch. One thing we could tighten later: `getTransactionReceipt` here runs without a `chainId`, so wagmi resolves it against whatever chain the wallet is currently on. So if someone submits a proposal on chain A and then reloads while their wallet happens to be on chain B, we won't find the receipt and we'll keep the record as in-flight — i.e. the same stale warning we're killing could sneak back in the cross-chain case.

Since we just freed up room by dropping `network`, we could stash a `chainId` on the persisted SUBMITTED record and pass it through:
```ts
getTransactionReceipt(wagmiConfig, { hash: state.hash, chainId: state.chainId })
```
No worries at all for this PR though — reconcile is best-effort and Return / Publish-anyway already rescue the user. Just flagging it so it's on the radar. 🙂

src/shared/utils/pendingTransactionManager/pendingTransactionManager.ts
```diff
@@ -141,11 +144,7 @@ export class PendingTransactionManager {
         intentId: string,
         state: IPendingTransactionState,
     ): void => {
-        // A PENDING update is a this-session send, so its outcome is awaited here (not interrupted).
-        if (state.status === PendingTransactionStatus.PENDING) {
-            this.liveSends.add(intentId);
-        }
-        // Merge retained meta so type/scope survive every status transition (and the persisted mirror).
+        // Merge retained meta so type/scope survive every status transition and the persisted mirror.
```
thekidnamedkd, line 147: Tiny nit: this comment still lists `network`, but that field got dropped in this PR — we could trim it so it doesn't send future readers hunting for a field that no longer exists:
```suggestion
        // Merge retained meta so type/scope/label survive every status transition and the mirror.
```

thekidnamedkd (approved):
Really nice fix — the root-cause framing (PENDING ghosts + orphaned SUBMITTED with no clear path) is spot on, and persist-only-SUBMITTED + reconcile-on-hydrate is a clean way to kill the ghost class. The `match`-based resume state machine and the details-with-Return warning are a lovely readability win too, and the tests cover the new paths well. Approving — a couple of non-blocking thoughts inline, both fine as follow-ups.

PR #1206, opened 2026-07-03
feat(APP-922): Refresh proposal card status in tx dialog after indexing

evanaronson:
# Description

- All proposal-card tx dialogs now use actual live data and the proposal status utility to populate the status in the card. 
- The status is refreshed after the new tx is indexed.
- The proposal slug is added to newly created proposals once the backend has indexed it.

## Type of Change

## Developer Checklist:

## Review Checklist:


src/modules/governance/dialogs/executeDialog/executeDialog.test.tsx
```diff
@@ -0,0 +1,232 @@
...
+        ).not.toBeInTheDocument();
+
+        // The fetch must use the locally derived slug, not the absent callback slug.
+        const fetchedWithLocalSlug = useProposalBySlugSpy.mock.calls.some(
+            ([params]) => params.urlParams.slug === 'SLUG-1',
+        );
+        expect(fetchedWithLocalSlug).toBe(true);
+    });
+
+    it('derives the indexed status from the governance proposal status slot', () => {
```
milosh86: I would remove this test, it's an implementation detail. Previous test tests behaviour and that's enough.

src/modules/governance/dialogs/publishProposalDialog/publishProposalDialog.test.tsx
```diff
@@ -231,3 +239,233 @@ describe('<PublishProposalDialog /> component', () => {
...
+        ).toBeInTheDocument();
+        expect(
+            screen.queryByText(
+                modulesCopy.proposalDataListItemStatus.statusLabel.DRAFT,
+            ),
+        ).not.toBeInTheDocument();
+        expect(screen.getByText('MY-PROPOSAL')).toBeInTheDocument();
+    });
+
+    it('derives the indexed status from the governance proposal status slot', () => {
```
milosh86: same comment as for execute dialog test ... it's implementation detail.

src/modules/governance/hooks/useIndexedProposalStatus/useIndexedProposalStatus.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const hasFreshIndexedProposal = isFetchEnabled && isFetchedAfterMount;
+    const proposalForStatus = hasFreshIndexedProposal
+        ? indexedProposal
+        : undefined;
+    const indexedProposalStatus = useSlotSingleFunction<
+        IProposal,
+        ProposalStatus
+    >({
+        params: proposalForStatus!,
```
milosh86: hm, does this even work? ... check SPP for example, I suspect this could throw unhandled error because GOVERNANCE_PROCESS_PROPOSAL_STATUS handlers expect proposal to be present and here there is a case when it's undefined. We use `!` type enforcement only when we are sure that data is present, but here it would be wron.

src/plugins/sppPlugin/dialogs/sppAdvanceStageDialog/sppAdvanceStageDialog.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+        ).not.toBeInTheDocument();
+
+        // The fetch must use the locally derived slug, not the absent callback slug.
+        const fetchedWithLocalSlug = useProposalBySlugSpy.mock.calls.some(
+            ([params]) => params.urlParams.slug === 'SLUG-1',
+        );
+        expect(fetchedWithLocalSlug).toBe(true);
+    });
+
+    it('derives the indexed status from the governance proposal status slot', () => {
```
milosh86: implementation detail test here too.

src/shared/components/transactionDialog/transactionDialog.test.tsx
```diff
@@ -615,3 +617,168 @@ describe('<TransactionDialog /> component', () => {
...
+            rerender(createTestComponent(nextProps));
+        });
+
+        expect(onIndexed).toHaveBeenCalledTimes(1);
+        expect(onIndexedNext).not.toHaveBeenCalled();
+    });
+
+    // The indexing step indicator legitimately changes from pending to success, so
+    // the assertion scopes to the consumer-rendered card region.
+    it('keeps children unchanged when onIndexed is omitted after indexing completes', () => {
```
milosh86: not sure what this test exactly validates? I guess it should just check it doesn't throw error when `onIndexed` is omitted? children is unchanged in any case, no?

src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -65,6 +66,15 @@ export const TransactionDialog = <TCustomStepId extends string>(
     // Make the onSuccess property stable to only trigger it once on transaction success
     const onSuccessRef = useRef(onSuccess);
 
+    // Track the latest onIndexed callback and whether it has already fired to trigger it exactly
+    // once when the transaction is indexed, regardless of callback or status-object identity changes
+    const onIndexedRef = useRef(onIndexed);
```
milosh86: This is BS which AI does all the time. `useRef` is not needed, just call `onIndexed` directly. `onIndexedFiredRef` makes sure it's called once.

milosh86 (commented):
Left a couple of comments.

src/modules/governance/dialogs/executeDialog/executeDialog.test.tsx
thekidnamedkd: Removed — the behaviour test right above already covers the status switch.

src/modules/governance/dialogs/publishProposalDialog/publishProposalDialog.test.tsx
thekidnamedkd: Removed here too — behaviour test above covers it.

src/plugins/sppPlugin/dialogs/sppAdvanceStageDialog/sppAdvanceStageDialog.test.tsx
thekidnamedkd: Removed — same as the others, behaviour test above is enough.

src/modules/governance/hooks/useIndexedProposalStatus/useIndexedProposalStatus.ts
thekidnamedkd: Good catch on the `!` — it was misleading. It doesn't actually throw though: when there's no fresh indexed proposal, `pluginId` is `''`, so `useSlotSingleFunction` finds no registered handler and never invokes anything (returns undefined → we fall back). The handler only ever runs once the proposal is present. Still, the `!` was claiming certainty we don't have there, so I've typed the param as `IProposal | undefined` and dropped the `!`.

src/shared/components/transactionDialog/transactionDialog.test.tsx
thekidnamedkd: Yeah, the children-unchanged assertion was trivial (children are passed straight through, so of course they don't change). Rewrote it to what actually matters: the dialog survives the indexed signal when `onIndexed` is omitted, i.e. the optional callback is a no-op.

src/shared/components/transactionDialog/transactionDialog.tsx
thekidnamedkd: Dropped the ref. `onIndexedFiredRef` already guarantees once, so `onIndexed` is now called directly and added to the effect deps — the fired-ref guard makes the identity change a no-op.

milosh86 (approved):
Nice, looks good now, thanks for addressing comments @thekidnamedkd 🙌

PR #1207, opened 2026-07-03
feat(APP-974): Integrate Vercel Speed Insights and Analytics

tyhonchik:
Installs and mounts Vercel's official App Router components so the [Speed Insights](https://vercel.com/aragon-app/app-next/speed-insights) and [Web Analytics](https://vercel.com/aragon-app/app-next/analytics) dashboards start receiving data on deployments. Previously both were empty because the collection packages were not wired into the app.

## Changes

- **Added dependencies** — `@vercel/analytics@2.0.1` and `@vercel/speed-insights@2.0.0`.
- **Mounted collectors** — `<Analytics />` and `<SpeedInsights />` (App Router variants) rendered inside `<body>` of the root layout (`layoutRoot.tsx`), after `<Providers>`. They run in `auto` mode and no-op outside Vercel/dev, so no env gating is needed.
- **Test isolation** — mocked both Vercel modules in `layoutRoot.test.tsx` (same pattern as the existing `Providers` mock) to avoid Jest ESM transform errors.

## CSP

No CSP changes required. In production/staging `'strict-dynamic'` propagates trust to the script the (nonce'd) app bundle injects; in preview/local `'self'`/`https:` cover the same-origin `/_vercel/.../script.js`; beacons hit same-origin `/_vercel/*` and are also allowed by `connect-src *`.

## Verification

- `pnpm type-check` — clean
- `pnpm test:changed` — 4 passed
- `ultracite`/lint-staged — clean

Post-deploy, confirm data lands on the Vercel dashboards (needs real traffic).

Closes APP-974

### DIFF
--- a/.changeset/app-974-vercel-insights.md
+++ b/.changeset/app-974-vercel-insights.md
@@ -1,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Integrate Vercel Speed Insights and Web Analytics to collect Web Vitals and page-view metrics on deployments.
--- a/package.json
+++ b/package.json
@@ -49,6 +49,8 @@
     "@sentry/nextjs": "^10.58.0",
     "@tanstack/react-query": "^5.101.0",
     "@tanstack/react-query-devtools": "^5.101.0",
+    "@vercel/analytics": "^2.0.1",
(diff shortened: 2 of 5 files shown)


milosh86 (approved):
👍

PR #1211, opened 2026-07-06
chore: restructure into a pnpm monorepo (apps/app)

tyhonchik:
# Description

This restructures the repository into a pnpm monorepo: the entire app moves into `apps/app`, and the root keeps only workspace tooling. Previously repo root == app root, so there was nowhere for the upcoming assistant service (APP-952) and shared packages to live. Now `apps/*` / `packages/*` are pnpm workspaces orchestrated by turbo, while the app itself is untouched — same `@aragon/app` name, version `1.33.0`, same scripts. The repo becomes a home for more than one deployable — before the wave of assistant/widget PRs lands.

**How to review:** the PR is exactly two commits. Commit 1 (`44744c20`) is a pure rename — 2684 files, `0 insertions, 0 deletions`, 100% R100 — skim it as a file list. Commit 2 is the real diff (~40 files of configs, CI and docs). Full history is preserved: `git log --follow` works across the move, and the rename commit is pinned in `.git-blame-ignore-revs`.

## Changes

- **Workspace root.** `pnpm-workspace.yaml` gains `packages: [apps/*, packages/*]` (overrides/settings unchanged); a new root `package.json` (private, no version) holds the tooling deps (turbo, biome, ultracite, husky, lint-staged, changesets, vercel) and proxy scripts via `turbo run …` / `pnpm --filter @aragon/app …` — so `pnpm type-check|lint|test|e2e|setup` keep working from the root and most CI needed no rewiring.
- **CI workflows regrouped per app** and made consistent: `app-release-v2` → `app-release-start`, `release-pr-*` → `app-release-pr-*`, `hotfix-start` → `app-hotfix-start`, `slot-docs` → `app-slot-docs`. Legacy `app-release.yml` (pre-release-PR direct flow) is deleted. Job names are unchanged, so the `test`/`deploy` required checks are unaffected.
- **Deploy needs no Vercel dashboard change.** Vercel never builds from git here — the GH job builds the artifact (`vercel build`) and uploads it as-is (`vercel deploy --prebuilt`). Those steps now run from `apps/app`; auto-link resolves the same project because the directory basename is still `app`. Old-layout branches keep deploying via their own copy of the workflow.
- **Release flow paths.** Versioning/changesets now target `apps/app`: `changeset version` writes `apps/app/CHANGELOG.md`, release workflows read the version via `npm-get-version-action` with `path: apps/app`, `read-changelog`/`changeset-version` actions updated accordingly.
- **Tooling configs.** biome globs, `.changeset/config.json` (`changedFilePatterns: apps/app/src/**`) and `.gitignore` point at `apps/app`; jest drops the agent-hook ignore patterns; guardrails scripts stay at the root; turbo gains `dev`/`build`/`lint:check`/`test:coverage` tasks.
- **Agent infra.** Rule-skill `globs` are prefixed with `apps/app/` (they match repo-root-relative paths); loader and adapter hooks are untouched. AGENTS.md gets a monorepo layout section; new root README.md.
- **MIGRATION.md** (temporary, root) — verified recipe for in-flight branches, see below.

## Note

- **Merge with a merge commit — do not squash or rebase.** `.git-blame-ignore-revs` pins the pure-rename commit hash; squashing would invalidate it and destroy the clean two-commit structure.
- A real git gotcha found while verifying the recipe (git 2.50, reproduced synthetically): on rebase across the move, edits to existing files relocate fine and files added *directly* to a renamed directory relocate with a `CONFLICT (file location)` notice — but **new files inside new subdirectories are silently left at the old path**. MIGRATION.md includes a mandatory post-rebase leftovers step that catches this; it was tested end-to-end (0 files lost).
- Per-app `paths` filters for workflows are deliberately *not* added — everything still runs on every PR. They'll come together with the assistant (required checks need the skipped-job gating pattern, not workflow-level `paths`).
- Verified locally: `pnpm install`, type-check, `lint:check`, jest 2010/2010, guardrails 30/30, production `next build`, dev-server boot (HTTP 200), YAML parse of all 29 workflow/action files, `git log --follow` back to the initial commit. Deploy/e2e/release flows are path-reviewed but will get their first real exercise in this PR's preview and the next release.

## Migration

For everyone with in-flight branches or local changes: follow [MIGRATION.md](../blob/chore/monorepo-skeleton/MIGRATION.md) — `git rebase origin/main`, then the leftovers one-liner. Local `.env` / `.env.local` move to `apps/app/` by hand. MIGRATION.md is temporary and will be deleted once pre-monorepo branches are merged.

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (commented):
Execution looks good. Splitting the pure rename from the tooling commit, pinning it in .git-blame-ignore-revs, and the MIGRATION.md recipe make this easy to follow, and preview + smoke are green. No blockers from me — the rest is just discussion since this is a skeleton.

On turbo: I think it's a reasonable choice. We're on Vercel and the assistant service plus shared packages are coming, so setting up the pattern now is fine. pnpm-only would cover what we actually do today (--filter and -r already handle parallel and topological runs) and Nx would be overkill, so no real objection to turbo itself.

The main thing is turbo isn't doing much for us yet. It's a single-package graph, and there's no remote cache wired (no TURBO_TOKEN/TURBO_TEAM in CI), so every CI run is a cold turbo run and the caching only helps locally right now. Fine for a skeleton, but the payoff only shows up once we have interdependent packages and remote caching on. I'd suggest turning on Vercel Remote Cache in a follow-up so CI actually benefits.

Two smaller things for whenever we next touch this. build has cache:false but declares .next/** outputs, so that task doesn't really do anything today (Vercel builds the artifact anyway) — worth either caching it or dropping it. And the CI split is actually right (graph tasks go through root/turbo, vercel and playwright run in apps/app directly) — a one-line comment there would keep it that way on purpose.

Good to merge as a skeleton. Only ask is that we track the remote-cache follow-up so turbo isn't running empty once the second package lands.

apps/app/package.json
```diff
@@ -1,131 +1,118 @@
 {
```
milosh86, line 1: double check if this re-formatting is expected

apps/app/tsconfig.json
```diff
@@ -1,64 +1,64 @@
 {
```
milosh86, line 1: same here

milosh86 (commented):
Great work here 🙌
Just left a comment to double check some suspicious formatting, but maybe it's just new biome version.

apps/app/package.json
tyhonchik: Not the biome version — the `!*.json` exclusion only matches root-level paths, so after the move to `apps/app/` biome started formatting these. Removed the blanket `!*.json`/`!*.jsonc` so all JSON is 4-space under one rule (locales/changeset/manifest stay excluded). No-op on content — everything was already 4-space.

milosh86 (approved):
Looks good 👍

asciiman (approved):
LGTM!

PR #1212, opened 2026-07-07
fix: update stage status logic to return accepted for executed proposals past max advance date

harryburger:
# Description

Noticed that an SPP proposal that was already executed can still show its last stage as "expired" / "Proposal not executed in time". We can see it in [here](https://app.aragon.org/dao/ethereum-mainnet/0xf204245b0B05E9A0780761E326552A569c1D6ceb/proposals/CIPX-3): the proposal executed fine back in May, but opening it in July shows the veto stage as expired.

Turns out this is purely a frontend thing. `getStageStatus` decides a stage is expired just by checking `now > maxAdvanceDate`:

```ts
const isExpired =
    !isSignalling &&
    stageIndex === currentStage &&
    maxAdvanceDate != null &&
    now > maxAdvanceDate;
```

It never looks at whether the proposal was actually executed. So once enough real time passes, `now` is always past the max-advance date and the stage flips to expired — even though the proposal was executed well before that deadline. That's why it looks fine one day and "expired" a few weeks later.

The backend data is correct the whole time (`executed.status: true`), and the overall proposal status already shows "Executed" — it's only the per-stage card that gets it wrong.

## Type of Change

## Developer Checklist:

## Tests

Added a case in `sppStageUtils.test.ts`: same setup as the existing "expired" test, but with `executed.status: true` — now expects `ACCEPTED` instead of `EXPIRED`. Whole suite passes (52/52).

### DIFF
--- a/.changeset/fix-spp-executed-stage-expired.md
+++ b/.changeset/fix-spp-executed-stage-expired.md
@@ -1,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Fix an executed SPP proposal showing its final stage as "expired" / "Proposal not executed in time". The per-stage status was derived from the clock alone (`now > maxAdvanceDate`) without checking whether the proposal had already been executed, so a proposal executed on time would flip to expired once viewed after its max-advance date. Executed proposals are now treated as accepted regardless of the max-advance date.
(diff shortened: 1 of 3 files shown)


thekidnamedkd (approved):
lgtm 💐

PR #1213, opened 2026-07-08
chore(APP-978): Update useSimulateProposalCreation params and return type (APP-947 1/4)

milosh86:
# Description

Updates `useSimulateProposalCreation` params and return type as part of standardizing dynamic proposal-creation permission checks.

- `network` param is now optional; the underlying `useCall` query is disabled until both a wallet address and a resolved `chainId` are available.
- The hook now returns a structured `IUseSimulateProposalCreationResult` (`isLoading`, `isError`, `result`) that distinguishes a **simulation failure** from a **failed simulation request**:
  - Only `ExecutionRevertedError` (walked from the wagmi `BaseError`) is treated as a simulation `failure`.
  - All other errors (e.g. RPC unavailable) surface as `isError` (request error), not a simulation result.
- Adds unit tests covering success, revert-as-failure, request errors, and disabled-query cases.

Relates to [APP-947](https://linear.app/aragon/issue/APP-947/dynamic-permission-checks-on-process-select-dialog) — **PR 1/4**.

Closes [APP-978](https://linear.app/aragon/issue/APP-978/update-usesimulateproposalcreation-params-and-return-type).

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (commented):
The error classification at the hook level is nice — distinguishing ExecutionRevertedError (user genuinely can't create) from RPC errors feels like the right abstraction.

A couple thoughts:

- **isLoading state isn't tested** — the mock always sets isLoading: false, so there's no assertion for { isLoading: true, isError: false, result: undefined } while the simulation is running. Not blocking, but would close a gap.
- **Nested ternary for the return** — the isSuccess / isSimulationError / else ternary works fine (isSuccess && isError can't both be true), but it might read cleaner as early returns or a small guard chain. Totally optional.

Everything else lines up cleanly with the SPP consumer update.

thekidnamedkd (commented):
The error classification at the hook level is nice — distinguishing ExecutionRevertedError (user genuinely can't create) from RPC errors feels like the right abstraction.

A couple thoughts:

- **isLoading state isn't tested** — the mock always sets isLoading to false, so there's no assertion for { isLoading: true, isError: false, result: undefined } while the simulation is running. Not blocking, but would close a gap.
- **The return value ternary** — works fine (isSuccess and isError can't both be true), but might read a bit cleaner as early returns or a guard chain. Totally up to you.

Only existing consumer is useSppPermissionCheckProposalCreation and the update there looks right. The optional network + enabled gating is a good improvement.

thekidnamedkd (approved):
Thanks for adding the isLoading test! On the return value — I played with a few shapes too and honestly nothing reads clearly better than what's there, so let's keep it. LGTM 🚀

PR #1214, opened 2026-07-08
feat(APP-979): Add proposal-creation simulation to ProcessDataListItem (APP-947 2/4)

milosh86:
# Description

Adds a dynamic proposal-creation eligibility check to `ProcessDataListItem`. Each process now simulates `createProposal` via `useSimulateProposalCreation`; when the simulation fails, the item is rendered disabled (non-clickable, dimmed) with a "Not eligible — View requirements" help text that links to the process's creation-eligibility settings section.

This is **2/4** of the PRs implementing **[APP-947](https://linear.app/aragon/issue/APP-947/dynamic-permission-checks-on-process-select-dialog)** (Dynamic permission checks on process select dialog). It builds on **1/4**, [#1213](https://github.com/aragon/app/pull/1213) (APP-978), and is therefore targeted at the `app-978-update-usesimulateproposalcreation-params-and-return-type` branch.

Resolves [APP-979](https://linear.app/aragon/issue/APP-979/add-simulation-to-processdatalistitem).

### Changes

- `ProcessDataListItem` gains `isDisabled` and `showNotEligibleHelpText` props: disabled items drop their `onClick`, get a neutral/dimmed style, and render an eligibility help text with a link to the requirements.
- New `SelectPluginDialogProcessListItem` wraps `ProcessDataListItem` and drives the disabled/help-text state from the `useSimulateProposalCreation` result; `SelectPluginDialog` now renders it instead of `ProcessDataListItem` directly.
- `DaoProcessDetailsPageClient` adds a `proposal-creation-eligibility` anchor so the "View requirements" link can deep-link into the eligibility section.
- New i18n keys: `app.shared.processDataListItem.notEligible`, `app.shared.processDataListItem.viewRequirements`.

> **Note:** merge/review the base PR (#1213) first — the "Files changed" diff will be clean once this is rebased onto the merged base.

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (commented):
The wrapper composition pattern here (SelectPluginDialogProcessListItem wrapping ProcessDataListItem) feels right — keeps the shared component unopinionated and the dialog-specific logic together.

One thing I'm curious about: the E2E smoke had 1 failed + 3 flaky tests. Is that something we can track down, or does it look CI-noise?

Otherwise the disabled semantics (aria-disabled + onClick undefined) are correct, and the help text link works even though it won't scroll to the section until PR 4/4 ships. No issue with that staging.

apps/app/src/shared/components/processDataListItem/processDataListItem.tsx
```diff
@@ -34,30 +49,52 @@ export const ProcessDataListItem: React.FC<IProcessDataListItemProps> = (
...
         <DataList.Item
             className={classNames('px-4 py-3 md:p-6', {
                 'border-primary-400 shadow-primary hover:border-primary-400 hover:shadow-primary':
                     isActive,
+                'border-neutral-100 bg-neutral-50': isDisabled,
             })}
             key={address}
             {...otherProps}
+            {...(isDisabled
+                ? { onClick: undefined, 'aria-disabled': true }
```
thekidnamedkd: One small a11y concern here: this disables the whole `DataList.Item`, but when `showNotEligibleHelpText` is true the item still contains the active “View requirements” link below.

Could we scope the disabled semantics to just the selectable row content, or render the help text/link outside the disabled subtree? The process row should be disabled, but the requirements link is the recovery action and should still be exposed as enabled.

milosh86: True, good catch. I'll try to improve it.

milosh86: Actually `aria-disabled` was wrong here, because of the way DataListItem is implemented in ui-kit.

_aria-disabled is not a global ARIA attribute — it's only valid on widget roles (button, link, option, group, …). On a role-less div it's invalid ARIA; screen readers ignore it at best._

I just removed it.

thekidnamedkd (dismissed):
Nice — the aria-disabled removal is actually the better fix than what I suggested. You're right that it's a widget attribute, and since DataListItem renders a role-less non-focusable div when onClick is undefined, there's no disabled subtree at all anymore — the requirements link stays a real link. Better all around. And good that the flaky e2e was just the backend being down. LGTM 🚀

thekidnamedkd (approved):
👌🤘

PR #1215, opened 2026-07-09
fix(APP-958): prevent app crash on wallet disconnect during transaction dialogs

harryburger:
# Description

Fixes the application crashing when the wallet disconnects while a transaction dialog is open (APP-958).

Wallet-requiring dialogs (vote, execute, publish proposal, token approve, …) assert a connected address during render. Disconnecting the wallet from the wallet itself while one of them was open made the assertion throw during render, and since `DialogRoot` renders outside the page-level error boundary, the throw unmounted the entire application.

Changes:

- Add a `requiresWallet` flag to `IDialogComponentDefinitions` and set it on the 15 dialogs carrying the connected-wallet assertion
- `DialogRoot` unmounts flagged dialogs as soon as the wallet address is gone — before they can re-render and throw — and removes them from the dialog stack once the disconnect settles. Connector switches are unaffected: while reconnecting, dialogs are only hidden and restored once the address returns
- Dialogs opened without a wallet (e.g. connect-wallet) are untouched, so the user can always reconnect
- Add a guardrail test that fails when a dialog carries the connected-wallet assertion but its definition lacks the flag, so new dialogs cannot reintroduce the crash

## Type of Change

## Developer Checklist:

## Review Checklist:


apps/app/src/shared/components/dialogRoot/dialogRoot.tsx
```diff
@@ -23,6 +25,27 @@ export const DialogRoot: React.FC<IDialogRootProps> = (props) => {
...
     const { t } = useTranslations();
     const { locations, close } = useDialogContext();
+    const { address, isReconnecting } = useWalletAccount();
+
+    // Wallet-requiring dialogs assert the address during render, so they must be unmounted from
+    // here before a disconnect makes them throw.
+    // Closing waits out reconnects, as connector switches blank the address for a moment.
+
+    const missingAddress = address == null;
+    const walletDisconnected = missingAddress && !isReconnecting;
```
thekidnamedkd: This is the one thing I'd fix before merge: per wagmi's `GetConnectionReturnType` (v3.5.1), `isReconnecting` is only true for `status: 'reconnecting'` — the mount-time `reconnectOnMount` path. A mid-session connector switch goes through `status: 'connecting'`, where `address` can be undefined while `isReconnecting` is false — so `walletDisconnected` flips true and we close the dialog instead of waiting it out. And at mount time (the only state this grace actually covers) the dialog stack is always empty, so the branch can't protect anything as written.

Suggest also excluding `isConnecting` (or gating on wagmi's true `isDisconnected`) — probably means exposing one more field from `useWalletAccount`. The render guard on `missingAddress` is right as-is; crash protection is unaffected either way. Alternatively, if closing on connector switch is behavior we're fine with, let's just correct this comment and the changeset's "Connector switches are unaffected" sentence so the changelog stays accurate.

apps/app/src/shared/components/dialogProvider/dialogProvider.api.ts
```diff
@@ -103,4 +103,9 @@ export interface IDialogComponentDefinitions<
      * Component to be rendered.
      */
     Component: ComponentType<IDialogComponentProps<TParams>>;
+    /**
+     * Marks a dialog that asserts a connected wallet during render. DialogRoot unmounts and
+     * closes it on wallet disconnect, before the assertion can throw and crash the application.
+     */
+    requiresWallet?: boolean;
```
thekidnamedkd, line 112: Non-blocking: the JSDoc says "asserts a connected wallet during render", but 10 of the 15 flagged dialogs don't actually assert — the flag we're really shipping is "needs a wallet to be meaningful". Suggest wording it as the behavioral contract ("requires a connected wallet; DialogRoot unmounts and closes it on wallet disconnect") plus one sentence that children stacked on a flagged dialog should also be flagged, so a future flagged-parent/unflagged-child stack doesn't orphan the child.

Related consistency question, also non-blocking: a few wallet-dependent transaction dialogs are left unflagged and will stay open on disconnect — VOTE_GAUGES_TRANSACTION, EXIT_QUEUE_WITHDRAW_TRANSACTION, spp ADVANCE_STAGE, EXECUTE_ACTIONS. Intentional, or worth flagging in the same pass?

.changeset/dialog-wallet-disconnect-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Fix the application crashing when the wallet disconnects while a transaction dialog is open. Wallet-requiring dialogs assert a connected address during render, so a disconnect made them throw with no error boundary above the dialog layer, unmounting the whole application. Dialog definitions now carry a `requiresWallet` flag and `DialogRoot` unmounts and closes flagged dialogs on disconnect before they can re-render without an address. Connector switches are unaffected: while reconnecting, dialogs are only hidden and restored once the address returns.
```
thekidnamedkd: Non-blocking wording nit: "restored once the address returns" is only true for the stack entry + params — the component unmounts while hidden, so internal state (form inputs, tx step) resets on remount. Fine trade-off, just worth softening here since this ships to the changelog (and the connector-switch sentence may need the same touch-up depending on how we resolve the `isReconnecting` thread).

thekidnamedkd (commented):
Really nice fix overall — the unmount-before-throw ordering is race-free (we re-render DialogRoot before the dialog child on the same wagmi store update), the integration test reproducing the actual invariant crash is great, and I checked the flag coverage: all 5 dialogs with render-time address invariants are flagged, and the unflagged `address != null` matches elsewhere are benign. One thing we should fix before merge (the wagmi state-model comment inline), the rest is non-blocking polish.

Non-blocking extras:
- Test gaps we could close cheaply: reconnect-restore (address returns → dialog re-renders, close never called), multi-dialog stack (flagged top closes, unflagged below becomes visible), and open-while-already-disconnected.
- Follow-up ticket material rather than this PR: DialogRoot still renders outside the page-level error boundary, so the flag only protects dialogs someone remembered to flag — an error boundary around the dialog layer would kill the whole crash class. Might also be a good rule-skill ("dialog asserts wallet ⇒ set requiresWallet").

milosh86 (commented):
Hey @harryburger, it looks good!

What I mentioned today, I was thinking about wrapping `ActiveDialogComponent` in an `ErrorBoundary` which would catch all errors, log em, and close the dialog. After thinking more about it, I think it's better to start with your more focused approach and we can consider this more general approach if needed later.

One thing only regarding your approach. Since now one must know about `requiresWallet` flag when creating a new dialog, let's add something like this to AGENTS.md: "When creating a new dialog which uses `TransactionDialog` internally pass `requiresWallet` flag to dialog definition.

apps/app/src/shared/components/dialogProvider/dialogProvider.api.ts
harryburger: Thanks for noting on this. 
Flagged all four in the same pass for consistency. Worth noting none of them could actually crash. `ExecuteActionsDialog` and `SppAdvanceStageDialog` read the address but never invariant on it, so this is about consistent behaviour rather than crash coverage. To maintain consistency I added on most of the dialogs that trigger transaction.

milosh86 (approved):
Looks good 👍

PR #1217, opened 2026-07-09
feat(APP-982): Manage process order and selection from select dialog (APP-947 3/4)

milosh86:
# Description

Lifts proposal-creation eligibility handling up to the `SelectPluginDialog` so process ordering and selection are managed globally from the simulation results (building on PR 2/4 #1214, which added per-item simulation to `ProcessDataListItem`).

- Each process reports its proposal-creation eligibility up to the dialog once its simulation resolves.
- Processes the user cannot create proposals in are sorted to the bottom of the list.
- While results are still loading, a spinner is shown and the list is hidden to avoid layout shift; the confirm action stays disabled until all results are ready.
- A preselected process the user is not eligible for is automatically deselected.

Relates to [APP-947](https://linear.app/aragon/issue/APP-947/dynamic-permission-checks-on-process-select-dialog) — **PR 3/4**.

Closes [APP-982](https://linear.app/aragon/issue/APP-982/manage-processes-order-and-selection-globally-from-select-dialog-based).

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (commented):
Lifting eligibility state to the dialog is the right call for global ordering. The boolean sort (eligible first) is clean and stable.

A couple things I'm curious about:

- **Spinner hiding the whole list** — full hide + spinner avoids flicker but loses context of what's loading. No strong opinion, but worth noting that showing the list items with a dimmed/loading state would give the user more context. Up to you whether that tradeoff is worth it here.
- **Multiple useEffect fires** — each item fires onEligibilityResult independently as its simulation resolves. The current === isEligible guard prevents churn so there's no actual bug, just noting that each resolve causes a re-render. With typically 5-10 plugins this is totally fine, but it's something to keep in mind if the list grows.

The auto-deselect logic is sound.

apps/app/src/modules/governance/dialogs/selectPluginDialog/selectPluginDialogProcessListItem.tsx
```diff
@@ -1,26 +1,46 @@
...
-    const { result } = useSimulateProposalCreation({
+    const { result, isLoading } = useSimulateProposalCreation({
         plugin: process,
         network: dao?.network,
     });
     const simulationFailed = result === 'failure';
 
+    useEffect(() => {
+        if (!isLoading) {
+            onEligibilityResult(pluginId, !simulationFailed);
```
thekidnamedkd, line 39: I think we should be a bit stricter before reporting the process as eligible here. `useSimulateProposalCreation` can return `result: undefined` when the call is disabled or when we get a non-revert request error, but this maps every non-loading/non-failure state to `true`.

Could we only report once there is a concrete simulation result (`success` or `failure`), or pass an explicit unknown/error state up to the dialog? Otherwise the parent can mark the item ready/eligible and enable selection even though the permission check did not actually produce an eligibility result.

milosh86: yeah, I was considering this, but I did some practical tests in different environments, i.e. with slow connections, and I found this as a sweet spot between simplicity, UX and correctness. 

This is more an UX improvement than a line of defence, so I was thinking that, in rare cases when RPC request fails, it would be better to fall back to existing create proposal guard (if clicked) than to introduce new UI to handle RPC failure, where basically there is no meaningful content to present.

milosh86: added a comment to make the intention and reasoning behind it more clear.

thekidnamedkd: Makes sense — checked that usePermissionCheckGuard is still wired into the create proposal page, so worst case on an RPC failure we just degrade to the existing guard flow. The comment you added captures it well. Settled 👍

thekidnamedkd (approved):
‼️

PR #766, opened 2025-09-17
feat(APP-4547): Enable Safe bodies to create proposals

milosh86:
# Description

Enable Safe bodies to create proposals.

- Update governance designer to allow Safe bodies to create proposals.
- Update governance designer to distinguish Safe external bodies - show Safe icon.
- Update permissions guard logic to check Safe bodies too.
  - Currently implemented by simulating proposal creation for SPP plugins.

Items not done in this PR:
- Permission guard currently checks correctly if user has a permission to create a proposal (by running simulation), but in the case user doesn't have the permission, in the list of "creation requirements" we don't currently list Safe bodies, because there is no easy way to know if body had `canCreateProposal` checked.
- Process details page does not show brand icon on Safe bodies.
- Process details page does not show Safe bodies in "proposal creation eligibility" section

Task: [APP-4547](https://aragonassociation.atlassian.net/browse/APP-4547)

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Developer Checklist:

- [x] Manually smoke tested the functionality locally
- [x] Confirmed there are no new warnings or errors in the browser console
- [x] (For User Stories only) Double-checked that all Acceptance Criteria are satisfied
- [x] Confirmed there are no new warnings on automated tests
- [x] Merged and published any dependent changes in downstream modules
- [x] Selected the correct base branch
- [x] Commented the code in hard-to-understand areas
- [x] Followed the code style guidelines of this project
- [x] Reviewed that the Files Changed in Github’s UI reflect my intended changes
- [x] Confirmed the pipeline checks are not failing

## Review Checklist:

- [ ] Tested locally that all Acceptance Criteria or Expected Outcomes are satisfied
- [ ] Confirmed that changes follow the code style guidelines of this project

[APP-4547]: https://aragonassociation.atlassian.net/browse/APP-4547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1218, opened 2026-07-09
fix(APP-965): Derive SPP proposal creation restriction from sub-plugin rules

thekidnamedkd:
## Summary

Fixes APP-965: the process details page showed **"Unrestricted: anyone can create a proposal"** for any connected wallet that was *eligible* to create one, even when the SPP process has real creation rules.

## Root cause

APP-4547 (#766) moved the SPP proposal-creation guard to an RPC simulation (`useSimulateProposalCreation`, a `useCall` for the connected account) but left the guard's `isRestricted` derived from that wallet-scoped result:

```ts
isRestricted: !permissionGranted // permissionGranted = hasSimulationSucceeded
```

`isRestricted` is a **process-level** fact ("are there restrictions on the action" — see `IPermissionCheckGuardResult`; token derives it from `minProposerVotingPower > 0`, multisig from `onlyListed`), so an eligible wallet (simulation succeeds) flipped the whole process to "Unrestricted" in `PermissionsDefinitionList`. The negative path (ineligible wallet) rendered correctly, which is why this only showed for wallets that could create.

## Fix

`useSppPermissionCheckProposalCreation` now aggregates `isRestricted` from the sub-plugin creation rules (restricted when any stage body restricts creation — the same source the displayed `settings` already came from), while `hasPermission` keeps using the RPC simulation for the wallet-level gate.

## Testing

- New `useSppPermissionCheckProposalCreation.test.ts` (5 cases): reproduces the bug against the pre-fix hook (restricted process + eligible wallet → was rendering unrestricted) and covers restricted/unrestricted aggregation, `hasPermission` from the simulation, and loading composition.
- `type-check`, `lint:check`, and the jest suite pass; additionally verified red→green from a clean checkout in an isolated linux/arm64 sandbox (deny-by-default network, node 24) as part of the sealbox clean-room dogfood.
- Changeset included (`@aragon/app` patch).

Linear: [APP-965](https://linear.app/aragon/issue/APP-965/proposal-eligibility-shows-unrestricted-for-eligible-wallets)

### DIFF
--- a/apps/app/src/plugins/sppPlugin/hooks/useSppPermissionCheckProposalCreation/useSppPermissionCheckProposalCreation.ts
+++ b/apps/app/src/plugins/sppPlugin/hooks/useSppPermissionCheckProposalCreation/useSppPermissionCheckProposalCreation.ts
@@ -71,10 +71,16 @@ export const useSppPermissionCheckProposalCreation = (
         result?.isRestricted ? result.settings : [],
     );
 
+    // The simulation only checks the connected wallet's permission; whether the process itself is restricted comes
+    // from the sub-plugin creation rules.
+    const isRestricted = pluginProposalCreationGuardResults.some(
+        (result) => result?.isRestricted === true,
+    );
+
     return {
         hasPermission: permissionGranted,
         settings,
         isLoading,
-        isRestricted: !permissionGranted,
+        isRestricted,
     };
 };
--- a/.changeset/fix-spp-proposal-eligibility-unrestricted.md
+++ b/.changeset/fix-spp-proposal-eligibility-unrestricted.md
@@ -1,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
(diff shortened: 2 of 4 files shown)


apps/app/src/plugins/sppPlugin/hooks/useSppPermissionCheckProposalCreation/useSppPermissionCheckProposalCreation.ts
```diff
@@ -71,10 +71,16 @@ export const useSppPermissionCheckProposalCreation = (
         result?.isRestricted ? result.settings : [],
     );
 
+    // The simulation only checks the connected wallet's permission; whether the process itself is restricted comes
+    // from the sub-plugin creation rules.
+    const isRestricted = pluginProposalCreationGuardResults.some(
```
milosh86, line 76: we probably could've used `settings.length` as a proxy for `isRestricted` too, but I think your version reads better 👍

milosh86 (approved):
Nice one 👍

PR #1225, opened 2026-07-13
fix(APP-956): Allow SPP rule conditions with only existing conditions

milosh86:
# Description

Governance designer's SPP rule condition validation required at least one governing body to remain checked, which made it impossible to configure the rule condition using only an existing condition. This updates validation to require at least one body **or** an existing condition, and updates the transaction utility to correctly build rules when proposal creation is only granted through existing conditions.

Fixes [APP-956](https://linear.app/aragon/issue/APP-956/allow-spp-rule-conditions-with-only-existing-conditions)

## Type of Change

## Developer Checklist:

## Review Checklist:


apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
```diff
@@ -99,23 +111,18 @@ export const CreateProcessFormProposalCreation: React.FC<
             ),
             rules: {
                 validate: (value) =>
-                    value !== ANY_WALLET && !canBodiesCreateProposals
+                    value !== ANY_WALLET && !hasProposalCreationSources
```
thekidnamedkd, line 114: The util tests cover code that already worked; the gate here is the untested part. Worth a small RTL test: no eligible body + one existing condition passes, no body + no condition errors, removing the last condition brings the error back. Not blocking.

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
```diff
@@ -56,6 +56,14 @@ export const CreateProcessFormProposalCreation: React.FC<
         'existingProposalCreationConditions'
     >({ name: 'existingProposalCreationConditions' });
 
+    const showExistingConditions =
+        isExistingConditionEnabled && isAdvancedGovernance;
+    const showBasicNotSupported =
+        isExistingConditionEnabled && !isAdvancedGovernance;
+
+    const hasExistingConditions =
+        showExistingConditions && existingConditionFields.length > 0;
```
thekidnamedkd, line 65: `length > 0` counts rows, not valid addresses, so clicking Add clears the error before anything's typed. The card's own address validation still blocks submit, so it's cosmetic. Could `useWatch` the addresses and count valid ones if we want it exact.

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
```diff
@@ -99,23 +111,18 @@ export const CreateProcessFormProposalCreation: React.FC<
...
+                    value !== ANY_WALLET && !hasProposalCreationSources
                         ? createProposalsError
                         : undefined,
             },
             defaultValue: LISTED_BODIES,
         },
     );
 
-    // Trigger proposalCreationMode validation on allowed bodies selection change
+    // biome-ignore lint/correctness/useExhaustiveDependencies: Trigger proposalCreationMode validation on allowed bodies selection change
```
thekidnamedkd: Reason is copied from the old comment, so it looks like the dep is accidental. Reword so it's clear it's intentional:

```suggestion
    // biome-ignore lint/correctness/useExhaustiveDependencies: hasProposalCreationSources is a trigger-only dep; validation must re-run when it changes
```

thekidnamedkd (commented):
Clean fix, nothing blocking. A few small inline notes. The one I would actually push on is a component test for the validation gate, since that gate is what the ticket's edge-case ask covers and it's currently untested (the new tests only cover the util, which already worked).

Separate optional thought: buildUpdateRulesTransaction still encodes an updateRules tx if conditionAddresses ends up empty. Can't happen via the form today (the card validates the address), but a guard there would be self-defending if the gate ever regresses.

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
```diff
@@ -158,7 +165,6 @@ export const CreateProcessFormProposalCreation: React.FC<
                 {processBodies.map((body) => (
                     <PluginSingleComponent
                         body={body}
-                        disableCheckbox={processBodies.length === 1}
```
thekidnamedkd, line 161: Two optional follow-ups. This isn't gated on the existingProposalCreationCondition flag, so with the flag off a single-body process can now be unchecked too (it shows the error instead of locking). Worth confirming that's intended. Also nothing passes disableCheckbox truthy anymore, so the prop and its branches are dead and can be cleaned up later.

milosh86: yeah, I intentionally removed this, because now error handling is fixed to be real-time and imo it's not worth extending this logic to support existing bodies.

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
milosh86: Added tests, good point.

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
milosh86: yeah, sound better this way

apps/app/src/modules/createDao/components/createProcessForm/createProcessFormProposalCreation/createProcessFormProposalCreation.tsx
milosh86: As I see, "valid address" error is not cleared when new condition is added, but regardless, you can't move on and you would see the error anyway. I don't think anything is confusing in the flow, so I wouldn't complicate it further.

thekidnamedkd (approved):
LGTM 🦋

PR #1233, opened 2026-07-14
fix(ci): scope app-finalize changeset guard; assistant release on-demand

tyhonchik:
## Why

The [`@aragon/app@1.33.3` finalize run](https://github.com/aragon/app/actions/runs/29327044080) failed at the **"Guard: no new changesets at finalize"** step, so the release was never tagged and production never deployed. Recovered manually (tag + Release on the tested SHA), but the guard is broken for the monorepo and will block **every** app release until fixed.

Root cause: the guard checked for *any* `.changeset/*.md`. In the monorepo, assistant changesets (`assistant-scaffold.md`, `assistant-contracts-scaffold.md`) live in `.changeset/` permanently — they're consumed by the assistant's own release flow, not the app ceremony. They tripped a guard that's only meant to catch *new app* changesets.

## Changes

- **Scope the finalize guard to `@aragon/app`** — parses each changeset's frontmatter and only fails on ones that bump `"@aragon/app"`, mirroring the `--ignore` scoping already used in [`changeset-version`](.github/actions/changeset-version/action.yml). Other workspaces' changesets no longer trip it. Verified against the current `.changeset/` contents: flags the app changeset, ignores both assistant ones.
- **Assistant Version-PR bot → `workflow_dispatch`** — was firing on every push to `main`, auto-opening a "Version Packages: assistant" PR. Now on-demand (trunk-based), matching the change already staged on the p4 feature branch.

### DIFF
--- a/.github/workflows/app-release-pr-finalize.yml
+++ b/.github/workflows/app-release-pr-finalize.yml
@@ -48,13 +48,19 @@ jobs:
           fetch-depth: 0
           token: ${{ steps.secrets.outputs.ARABOT_PAT }}
 
-      - name: "Guard: no new changesets at finalize"
+      - name: "Guard: no new @aragon/app changesets at finalize"
         shell: bash
         run: |
           set -euo pipefail
-          if ls .changeset/*.md >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


milosh86 (approved):
👍

PR #1235, opened 2026-07-14
feat(APP-266): in-app support chat with Linear intake

tyhonchik:
Replaces the external support-portal link in the footer with an in-app support chat that turns a natural conversation into a well-formed Linear ticket. The chat runs against a dedicated assistant service (`apps/assistant`, deployed separately) and is gated behind the `supportChat` feature flag (on for local/dev/preview, off for staging/production), so this can ship dark and be enabled per environment. When the flag is off — or the service is unreachable — the footer still links to the existing support portal.

The whole feature lives in three workspaces: the `@aragon/assistant` Hono service (LLM intake pipeline → Linear), the `@aragon/assistant-chat` React widget, and the `@aragon/assistant-contracts` shared zod schemas.

## Changes

- **Assistant service (`apps/assistant`)** — deterministic intake pipeline (classify intent → extract fields → respond) over the AI Gateway, idempotent Linear issue creation, per-session file queue with client-direct Vercel Blob uploads + magic-byte validation, per-IP rate limits (RPM + sessions/day), and PII-safe Sentry observability.
- **Chat widget (`@aragon/assistant-chat`)** — side-drawer chat (~30vw / min 500px), streaming replies, attachment picker/drag-drop/paste, a single compact sticky "Ready to send" bar with collapsible details, request history, and a support-portal escape hatch on hard failures.
- **App integration (`apps/app`)** — footer entry behind the `supportChat` flag, lazy-loaded widget, health-gated fallback dialog, and silent debug-context collection (route, DAO/network, app version, wallet + chainId, recent transactions).
- **Contracts (`@aragon/assistant-contracts`)** — shared request/response/limits schemas consumed by both sides; built to `dist/` via Turbo `^build`.

## Conversation behaviour

- **Neutral, concise tone, no emoji.** Replies in the user's language; the assistant collects requests rather than answering product questions, and refuses those briefly without lecturing.
- **Gathers before closing.** Once it has enough, it restates its understanding and invites anything that helps debugging (steps, screenshot, logs) plus an optional email — instead of just pointing at the button.
- **File-aware.** The model is told which files are attached (names/count only, never contents), so it acknowledges them instead of claiming it can't see them.
- **English tickets.** The ticket + live preview (summary/description/steps) are always written in English (translated when needed); the email is kept verbatim.

## Context, debug & privacy

- The user never sees a raw context dump — a quiet "Debug info is attached automatically" line stands in.
- The Linear ticket carries route, DAO, network, chainId, wallet, recent transactions, and a Sentry `user.id:<wallet>` pointer for replay lookup.
- Attachments travel through a private path (Blob → private Linear queue), gated by type/size allowlist and rate limits.

## Models & limits

- Cheap, fast, recent models from proven providers with strong English + tool-calling: **primary `openai/gpt-5-nano`**, **fallback `google/gemini-2.5-flash-lite`** (both ≤ $0.10/M input; the Gateway retries on the fallback automatically).
- Per-session limits: 8 turns, 5000 token budget, 3 files × 5 MB; per-IP 5 rpm / 5 sessions per day. All env-overridable.

## Testing

- `pnpm type-check`, `pnpm lint:check`, `pnpm test` green across workspaces (assistant, widget, contracts, app).
- Live model behaviour is validated by the assistant LLM smoke checks (nightly + post-deploy).

## Deferred (documented, not in this PR)

- Console-log ring buffer attached as a scrubbed `.log`.
- Automated vision moderation of uploaded images at `/files/confirm`; CSAM/legal handling needs a dedicated provider.
- Vercel BotID Deep Analysis + WAF Attack Challenge Mode for distributed-bot defense (needs the assistant project on Pro + a spike).


apps/app/src/modules/application/components/supportChat/supportChat.tsx
```diff
@@ -0,0 +1,44 @@
...
+// front of the drawer. Service failures surface inside the widget, where the support portal
+// stays one click away (header link + error escape hatches).
+export const SupportChat: React.FC<ISupportChatProps> = (props) => {
+    const { isOpen, onClose } = props;
+
+    const appContext = useSupportAppContext();
+
+    // Mount the widget on first open and keep it mounted so the conversation survives closing
+    // and reopening the drawer.
+    const [hasOpened, setHasOpened] = useState(false);
```
milosh86, line 22: `AssistantChatLazy` has isOpen and `SupportChat` is lazy loaded, what's the main reason behind this additional logic?

milosh86 (commented):
Looks solid, one small comment/question

apps/app/src/modules/application/components/supportChat/supportChat.tsx
tyhonchik: `isOpen` only toggles the drawer — it doesn't gate the lazy chunk. `next/dynamic` fetches the widget bundle on mount, so rendering `AssistantChatLazy` unconditionally would download it on every page load. `hasOpened` defers the first mount until the first open, then keeps it mounted so the conversation survives close/reopen.

milosh86 (approved):
👍

PR #1238, opened 2026-07-15
feat(APP-1003): Iterate on permissions graph view from feedback

thekidnamedkd:
# Description

This PR delivers the **APP-1003** refinement pass on top of the finished APP-942 graph view.

The underlying APP-942 permissions graph work is already checked in; this summary only covers the APP-1003 refinement delta.

Related base graph work: APP-942 / superseded graph PR #1202.

## APP-1003 refinement (permissions feedback)

- Consolidate the view into a single graph screen driven by toggles; remove the Granted/Other tabs
- Treat each permission as a distinct node instead of merging under a shared "anyone" header
- Standardize resource-link display and show the permission ID above the name
- Replace explicit "no condition"/"no description" labels with a dash placeholder
- Add graph full-screen mode with Escape-to-exit
- Rename the secondary switch to `Show supporting permissions` and add concise supporting-permissions help copy
- Add informational tooltips to the Who / Where permissions-list headers
- Prefer backend-enriched `who` / `where` / `conditionEntity` permission metadata and keep local plugin/account matching as a fallback
- Move DAO self-permission stacks south of the DAO node and align stack handles
- Hide stale/supporting entities from the default graph and explicitly label historical/uninstalled plugin nodes when shown

## Type of Change

## Developer Checklist:

- [x] Manually smoke tested the functionality locally against the Citrea DAO permissions path above
- [ ] Confirmed the pipeline checks are not failing — pending CI

## Review Checklist:

- [ ] Tested in a preview or locally that the graph and list permissions views satisfy the APP-942 / APP-1003 UX paths


apps/app/src/modules/settings/components/executeSelectorConditionSlot/executeSelectorConditionSlot.tsx
```diff
@@ -1,60 +1,255 @@
...
+                            </Link>
+                        )}
+                    </div>
+                </div>
+            ))}
+        </div>
+    );
+};
+
+const AllowedActionsSkeleton: React.FC = () => (
```
milosh86: Let's split this file, it seems to contain too many different components and utils.

apps/app/src/modules/settings/components/executeSelectorConditionSlot/executeSelectorConditionSlot.tsx
```diff
@@ -1,60 +1,255 @@
...
+        network,
+        pluginAddress,
+    } = props as IExecuteSelectorConditionSlotProps;
     const { t } = useTranslations();
 
-    const allowedActions = toAllowedActions(selectors, targets);
-    const hasAllowedActions = allowedActions.length > 0;
+    const rawAllowedActions = toAllowedActions(selectors, targets);
+    const hasRawAllowedActions = rawAllowedActions.length > 0;
+    const shouldFetchDecodedActions =
```
milosh86: Maybe `shouldShowDecodedActions` reads better? ... I expect `shouldFetchDecodedActions` to be used in a query or something like that.

apps/app/src/modules/settings/components/membershipConditionSlot/membershipConditionSlot.tsx
```diff
@@ -11,10 +11,12 @@ import { useTranslations } from '@/shared/components/translationsProvider';
  * "Member of multisig".
  */
 export const MembershipConditionSlot: React.FC<IConditionData> = (props) => {
-    const { onlyListed } = props;
+    const { onlyListed, minApprovals } = props;
     const { t } = useTranslations();
 
     const isMemberGated = onlyListed === true;
+    const approvalsLabel =
```
milosh86: can we just render `minApprovals` directly? It's either undefined or number anyway, no?

apps/app/src/modules/settings/components/permissionsGraph/permissionsGraph.tsx
```diff
@@ -0,0 +1,371 @@
...
+
+interface IPermissionDetailPanelProps {
+    chainId?: number;
+    edge: IPermissionGraphEdge;
+    network?: IDao['network'];
+    nodes: IPermissionGraph['nodes'];
+    onClose: () => void;
+}
+
+const PermissionDetailPanel: React.FC<IPermissionDetailPanelProps> = ({
```
milosh86: This component deserves probably its own file

milosh86 (commented):
Looks great, left some housekeeping comments.

apps/app/src/modules/settings/components/executeSelectorConditionSlot/executeSelectorConditionSlot.tsx
thekidnamedkd: Done in f68e2e813. Renamed the local boolean to shouldShowDecodedActions so it describes the render branch instead of sounding like the query trigger.

apps/app/src/modules/settings/components/executeSelectorConditionSlot/executeSelectorConditionSlot.tsx
thekidnamedkd: Done in f68e2e813. Split the execute-selector condition slot into smaller files: raw/decoded list rendering, decoded-loading orchestration, and shared conversion/filtering utilities. The entry component now only coordinates the slot state.

apps/app/src/modules/settings/components/membershipConditionSlot/membershipConditionSlot.tsx
thekidnamedkd: Done in f68e2e813. The component now renders minApprovals directly when present; I added a narrow local condition payload type because the shared condition slot payload is typed as unknown at the boundary.

apps/app/src/modules/settings/components/permissionsGraph/permissionsGraph.tsx
thekidnamedkd: Done in f68e2e813. Moved PermissionDetailPanel into permissionsGraph/permissionDetailPanel.tsx and left permissionsGraph.tsx focused on graph state/render orchestration.

milosh86 (dismissed):
Great job 🙌 I've just tested it and it looks great!

One usage feedback, not related to implementation: `Unknown address` is not really helpful. You need to click to see which address is that, but it could be handy to have truncated address as a label instead.

PR #1240, opened 2026-07-16
ci(APP-997): auto-start weekly app release every Monday 6PM CET

tyhonchik:
# Description

Adds an auto release: **App Release Start** now also runs on a schedule — every Monday at 6PM CET (`0 17 * * 1` UTC) — so the weekly release no longer needs a manual kick-off. Manual dispatch keeps working exactly as before.

A new `schedule-gate` job (scheduled runs only, manual dispatch bypasses it) green-skips instead of starting a duplicate or empty release when:

- a `release/app/*` PR is already open;
- there are no pending changesets for `@aragon/app`.

Docs: the Start Release step in `release-process.md` now documents the schedule.

Task: [APP-997](https://linear.app/aragon/issue/APP-997/auto-releases)

## Type of Change

- [x] CI/docs only — no version bump, no changeset needed

## Developer Checklist:

### DIFF
--- a/.github/workflows/app-release-start.yml
+++ b/.github/workflows/app-release-start.yml
@@ -1,6 +1,10 @@
 name: App Release Start
 
 on:
+  schedule:
+    # Auto release: every Monday at ~6PM CET (17:07 UTC). Off the top of the
+    # hour because GitHub delays or drops schedules queued at :00.
+    - cron: '7 17 * * 1'
   workflow_dispatch:
     inputs:
       base_commit:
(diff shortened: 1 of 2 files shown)


thekidnamedkd (commented):
Review notes:

1. `.github/workflows/app-release-start.yml` — the schedule is set to `0 17 * * 1`, i.e. the top of the hour. GitHub documents that scheduled workflows can be delayed or even dropped during high-load periods, especially at the start of every hour. Since this workflow is meant to reliably start the weekly release, I’d move it off minute `0`, e.g. `7 17 * * 1` or `17 17 * * 1`.

2. `.github/workflows/app-release-start.yml` — the duplicate-release gate uses `gh pr list` without an explicit limit. `gh pr list` defaults to 30 results, so if the repo ever has more than 30 open PRs to `main`, an existing `release/app/*` PR can be outside the fetched page and the scheduled run can miss it. Current repo count is 7, so this is not failing today, but I’d add `--limit 100` or use a more targeted search/head query.

Validation I ran locally:
- Rebasing the local checkout onto latest `origin/main` succeeded; no push performed.
- `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/app-release-start.yml` exited 0.
- YAML parse confirmed both `schedule` and `workflow_dispatch` triggers and the gate condition.
- `git diff --check origin/main...HEAD` exited 0.

.github/workflows/app-release-start.yml
```diff
@@ -1,6 +1,9 @@
 name: App Release Start
 
 on:
+  schedule:
+    # Auto release: every Monday at 6PM CET (5PM UTC).
+    - cron: '0 17 * * 1'
```
thekidnamedkd: Small reliability tweak: GitHub can delay or drop scheduled workflows around the top of the hour. Since this one is the weekly release kickoff, I’d move it off `:00`, something like `7 17 * * 1`.

tyhonchik: Done — moved to `7 17 * * 1` (17:07 UTC, ~6PM CET). Applied the same tweak to the backend PR (aragon/app-backend#1460): `37 16 * * 1`, off :30 for the same reason.

thekidnamedkd (approved):
lgtm 🐨

PR #1242, opened 2026-07-16
feat(APP-998): support chat side panel, email escape hatch, centralized copy

tyhonchik:
This turns the support chat from a blur overlay into a side panel the app resizes around, and centralizes all chat/assistant copy. Previously the chat was a Radix Dialog pinned over the page — backdrop blur, focus trap, page frozen behind it — opened from a footer link, with its strings scattered across a dozen widget components and the service. Now the chat is a full-height column next to the app (the header, content and footer shrink to fit), toggled from a button at the trailing edge of every navigation bar, and every string has exactly one home. The panel shell moves out of the widget into the app layout — the upcoming assistant-ui rework only swaps the panel's innards.

## Changes

- **Overlay → in-flow side panel.** On `lg+` the panel is a sticky, full-viewport-height column (`clamp(500px, 30vw, 640px)`) with a ~300ms width animation; below `lg` it opens fullscreen with the page scroll locked behind it. Non-modal by design: no backdrop, no focus trap, no ESC-close — the page stays scrollable and interactive while the chat is open, and `overscroll-contain` keeps the chat scroll from chaining into the page. While closed the panel stays mounted (the conversation survives reopening) but `inert`.
- **Trigger moves footer → header.** A new `SupportChatTrigger` (feedback icon ↔ chevron) sits at the trailing edge of all three navbars via a new `trailing` slot on `Navigation.Container`, so it lands right next to the panel. The footer "Support" entry reverts to the plain external portal link regardless of the flag; open/close state lives in a new `SupportChatContextProvider` in the root layout.
- **Support portal deprecated inside the chat.** The header link and both error escape hatches now point to `support@aragon.org` (mailto) instead of the portal; the `supportPortalUrl` prop is removed from the widget API end-to-end.
- **Copy centralized, one module per package.** Every user-facing widget string — including the service-error texts and file alerts — moves to `packages/assistant-chat/src/copy.ts`; components only reference it. Service-side, the static Linear ticket texts move from `linear/issueBody.ts` into `chat/prompts/issueTexts.ts`, so all service copy (prompts, fixed replies, ticket texts) is edited under `chat/prompts/`. Pure string moves, zero wording changes.
- **`ChatDrawer` deleted** along with the widget's `@radix-ui/react-dialog` dependency; `AssistantChat` now just fills whatever container the host renders it in.

## Note

The chat UI itself (message list, composer, status strip) is intentionally untouched — the assistant-ui evaluation happens in a follow-up branch based on this one. Wording changes beyond the portal→email swap are also out of scope here.


apps/app/src/modules/application/components/supportChat/supportChatPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+            ? classNames(
+                  'fixed inset-0 z-[var(--guk-dialog-content-z-index)] flex',
+                  'lg:inset-auto lg:z-auto lg:flex lg:border-neutral-100 lg:border-l lg:shadow-neutral-lg',
+                  panelWidthClassName,
+              )
+            : 'hidden lg:flex lg:w-0',
+    );
+
+    return (
+        <aside
```
thekidnamedkd, line 74: The closed panel is still an implicit `complementary` landmark on desktop because it renders as `lg:flex lg:w-0` while `inert`. That gives DAO pages two `complementary` landmarks and matches the smoke failures on `getByRole('complementary')`.

Can we hide the landmark while closed, for example by only setting the `aside`/`aria-label` when `isOpen`, or by using `aria-hidden={!isOpen}` in addition to `inert`?

apps/app/src/modules/application/components/supportChat/supportChatPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+        return null;
+    }
+
+    const panelClassNames = classNames(
+        'justify-end overflow-hidden bg-neutral-0',
+        'lg:sticky lg:top-0 lg:h-dvh lg:shrink-0 lg:self-start',
+        'lg:transition-[width] lg:duration-300 lg:ease-in-out',
+        isOpen
+            ? classNames(
+                  'fixed inset-0 z-[var(--guk-dialog-content-z-index)] flex',
```
thekidnamedkd, line 66: On mobile this fullscreen drawer covers the app and locks page scroll, but the app behind it stays in the tab order. I can open the chat at mobile width and still tab through Connect, DAO cards, footer links, etc. behind the panel.

Can we add mobile-only focus containment here? Desktop can stay non-modal, but below `lg` the covered app content should probably be inert until the drawer closes.

.github/release-scopes.yml
```diff
@@ -0,0 +1,13 @@
...
+# changesets, which breaks every scoped flow. CI enforces this via `pnpm validate:changesets`.
+# When a new workspace lands, add it to a scope HERE — flows and validation read this file.
+
+app:
+  - "@aragon/app"
+
+assistant:
+  - "@aragon/assistant"
+  - "@aragon/assistant-contracts"
+  - "@aragon/assistant-chat"
```
thekidnamedkd: `@aragon/assistant-chat` ships inside the app bundle (`@aragon/app` depends on it and the app release filter includes `packages/assistant-chat/**`), but this maps it only to the assistant release scope. That means an app release can deploy assistant-chat changes without consuming the assistant-chat changeset/version/changelog; the package version would move later through an assistant flow that doesn't deploy the widget.

Can we either put app-bundled packages in the app scope, or make the coupled release behavior explicit so the tag that deploys the app also consumes the widget changeset it ships?

thekidnamedkd (changes requested):
This is a really nice cleanup overall — the side-panel behavior lines up with APP-998, the old fixed-message copy now says \`Prepare ticket\`, and the earlier a11y threads look handled. One release-safety thing before this lands:

App releases now own \`@aragon/assistant-chat\` in \`.github/release-scopes.yml\`, but the app finalize guard still only blocks newly added \`@aragon/app\` changesets. So an \`@aragon/assistant-chat\` changeset could be added to an already-open app release PR after \`changeset version\`, then the finalize workflow would still tag the app release without consuming/versioning that widget changeset.

Could we update the finalize guard to check the whole \`app\` release scope — ideally by reusing \`releaseScopes.js\` — instead of matching only \`@aragon/app\`? That keeps the new scope ownership airtight.

thekidnamedkd (approved):
LGTM 🧃

PR #1252, opened 2026-07-21
chore(deps): dependency update — patch audit, prune overrides

tyhonchik:
Recurring dependency refresh (`pnpm up -r`) with an audit pass and an overrides review. Bumps the catalog + `apps/app` to current minors, moves `vercel` to the latest major, and re-optimizes the overrides so the audit drops from 12 vulnerabilities to 1 (the remaining one has no upstream patch). Two upgrades that would drag unrelated churn into this PR are deliberately held and tracked for a separate cleanup task.

## Changes

- **Dependencies**: `pnpm up -r` across all workspaces; only `apps/app` had non-catalog bumps (next, sentry, wagmi, viem, framer-motion, recharts, reown/walletconnect, etc.). `pnpm@11.13→11.15`. `vercel ^52 → ^56.1.0` (via `pnpm up vercel -r --latest`; the 7-day `minimumReleaseAge` cooldown caps it at 56.1.0 for now, floats to 56.4.x later).
- **Audit 12 → 1**:
  - `axios` override was stale (`<1.15.2`); the advisory range widened, so `axios@1.16.0` leaked through → widened to `axios@<1.18.0: '>=1.18.0'`.
  - vercel@56 spread `undici` across new dependency paths that the old path-scoped `@vercel/node>undici` pin didn't cover → replaced with a global floor `undici@<6.27.0: ^6.27.0` (keeps 6.x, leaves undici 7.x untouched).
  - Added `brace-expansion@<1.1.16: ^1.1.16` (caret, not `>=`, to avoid jumping the 1.x line to 5.x and breaking `minimatch@3`).
  - Verified every other override is still active (target present, floor holding); none were dead.
- **Remaining audit finding (1 low)**: `elliptic <=6.6.1` — patch `6.6.2` is not published yet, so it can't be overridden. Dev-only (synpress e2e → ethers). Known.

## Held for follow-up (see cleanup ticket)

- **ultracite pinned to exact `7.8.3`**: 7.9.4 tightened Biome rules → a 122-file auto-reformat + 8 manual lint fixes in `assistant-chat`. Out of scope for a deps bump.
- **`sanitize-html` held at 2.17.5** via `sanitize-html@>=2.17.6: 2.17.5`: 2.17.6 moved to `htmlparser2@^12` (ESM-only), which Jest can't parse without `transformIgnorePatterns` work.

## Verification

- `pnpm audit`: 1 low (elliptic, no upstream patch)
- `pnpm lint:check` ✅ · `pnpm type-check` ✅ · `pnpm test` (no e2e) ✅ **2164/2164**
- `pnpm install --frozen-lockfile` consistent

## Notes

- No changeset: the changesets config tracks `src/**` only, and these are root manifest/lockfile changes — deps ride the next release.
- `pnpm dedupe` still exits non-zero on a **pre-existing** `playwright-core` missing peer (synpress-cache pins 1.48.2 vs `@playwright/test@1.61`) — present on `main`, not introduced here; folded into the cleanup ticket rather than papered over with `ignoreMissing`.


thekidnamedkd (approved):
LGTM 🪁

PR #1253, opened 2026-07-22
feat(APP-1001): Filter plugin actions in action composer by DAO permissions

milosh86:
# Description

Filters OSx-related plugin actions out of the action builder when the DAO does not hold the required permission to execute them, so users are no longer offered actions that would fail on execution.

Each plugin action item can now declare a `requiredPermissionId` (keccak256 permission id). When building the DAO action list, `actionComposerUtils` keeps only the plugin items whose required permission is granted to the DAO, then prunes any plugin groups left without items.

Details:
- Add `requiredPermissionId?: string` to `IActionComposerInputItem`. Matching is by permission id and target contract (`defaultValue.to` as `where`); `who` is checked against DAO address.
- `filterItemsByPermission` fails closed while permissions are still loading (`undefined`), always keeps items without a `requiredPermissionId`, and throws (`invariant`) when a permission-tagged item has no resolvable target address (misconfiguration guard).
- `pruneEmptyGroups` drops plugin groups whose actions were all filtered out.
- Tag the relevant plugin actions with their permission ids:
  - Multisig: add/remove members and update settings → `UPDATE_MULTISIG_SETTINGS_PERMISSION`
  - Token: mint → `MINT_PERMISSION`, update voting settings → `UPDATE_VOTING_SETTINGS_PERMISSION`
  - Lock-to-vote: update settings → `UPDATE_SETTINGS_PERMISSION`
  - Metadata: set metadata → `SET_METADATA_PERMISSION`
- Add unit tests for `getDaoActions` covering permission filtering and empty-group pruning.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
+++ b/apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
@@ -77,7 +77,10 @@ class ActionComposerUtils {
      * Filters out plugin action items whose required OSx permission is not granted
      * to the DAO. Items without a `requiredPermissionId` are always kept, and
      * filtering is skipped entirely (fail-open) while `permissions` is undefined
-     * (not yet loaded). Matching is by permission id only, case-insensitively.
+     * (not yet loaded).
+     * A required permission counts as granted when the DAO's permission list
+     * contains a matching `permissionId` on the action's target contract
+     * (`defaultValue.to` as `where`), matched case-insensitively.
      *
      * @param items - Plugin action items to filter.
      * @param permissions - Permissions granted to the DAO, or undefined when not loaded.
@@ -91,17 +94,27 @@ class ActionComposerUtils {
             return items;
         }
 
-        const grantedIds = new Set(
-            permissions.map((permission) =>
-                permission.permissionId.toLowerCase(),
-            ),
-        );
+        return items.filter((item) => {
+            const { requiredPermissionId, defaultValue } = item;
 
-        return items.filter(
-            (item) =>
-                item.requiredPermissionId == null ||
-                grantedIds.has(item.requiredPermissionId.toLowerCase()),
-        );
+            // Items without a required permission are never permission-filtered.
+            if (requiredPermissionId == null) {
+                return true;
+            }
+
+            // Fail-open when the action's target contract cannot be resolved.
+            const where = defaultValue?.to;
+            if (!where) {
+                return true;
+            }
+
+            return permissions.some(
+                (permission) =>
+                    permission.permissionId.toLowerCase() ===
+                        requiredPermissionId.toLowerCase() &&
+                    addressUtils.isAddressEqual(permission.whereAddress, where),
+            );
+        });
     };
 
     /**
--- a/apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
+++ b/apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
@@ -8,6 +8,7 @@ import {
     actionViewRegistry,
 } from '@/shared/utils/actionViewRegistry';
 import { ipfsUtils } from '@/shared/utils/ipfsUtils';
+import { permissionNameUtils } from '@/shared/utils/permissionNameUtils';
 import { pluginRegistryUtils } from '@/shared/utils/pluginRegistryUtils';
 import {
     type IProposalAction,
@@ -32,6 +33,10 @@ import {
     type IGetNativeActionGroupsParams,
     type IGetNativeActionItemsParams,
 } from './actionComposerUtils.api';
+
+const setMetadataPermissionId = permissionNameUtils.getPermissionId(
+    'SET_METADATA_PERMISSION',
+);
 
 class ActionComposerUtils {
     // The TransferActionLocked is a UI-only variant of the TransferAction which locks the token field to the one set as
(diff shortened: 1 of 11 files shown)


apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
```diff
@@ -51,15 +56,83 @@ class ActionComposerUtils {
...
+                where != null && where !== '',
+                `filterItemsByPermission: action "${item.id}" declares a requiredPermissionId but has no target address.`,
+            );
+
+            return permissions.some(
+                (permission) =>
+                    permission.permissionId.toLowerCase() ===
+                        requiredPermissionId.toLowerCase() &&
+                    addressUtils.isAddressEqual(permission.whereAddress, where),
+            );
```
thekidnamedkd, line 126: We're matching `permissionId` + `where` only, but APP-1001 keys executability on the DAO holding the permission as `who`. So the Alchemix case (granted to someone, but not the DAO) still shows here and reverts on execution. Closing it means threading `dao.address` through and matching `whoAddress` too. Fine as a follow-up if we'd rather scope it out.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
```diff
@@ -51,15 +56,83 @@ class ActionComposerUtils {
...
+     * with no resolvable target throws, as that indicates a misconfigured action.
+     *
+     * @param items - Plugin action items to filter.
+     * @param permissions - Permissions granted to the DAO, or undefined when not loaded.
+     * @returns The items the DAO is authorized to execute.
+     */
+    private filterItemsByPermission = (
+        items: IActionComposerInputItem[],
+        permissions: IGetDaoActionsParams['permissions'] = [],
+    ): IActionComposerInputItem[] => {
```
thekidnamedkd, line 96: Minor, non-blocking: the doc says we fail open while loading, but the `= []` default fails closed, so gated actions blink out then back in as permissions page in. If we want fail-open, guarding on `== null` does it (`!length` would also hide them for a DAO with genuinely zero perms):

```suggestion
    private filterItemsByPermission = (
        items: IActionComposerInputItem[],
        permissions: IGetDaoActionsParams['permissions'],
    ): IActionComposerInputItem[] => {
        if (permissions == null) {
            return items;
        }
```

thekidnamedkd (dismissed):
Couple of inline notes, nothing gating from me. Main one's the `who` check APP-1001 asks for.

thekidnamedkd (commented):
Reviewed, leaving this as comments. Two notes inline: the `who` check APP-1001 wants (main one, happy for it to be a follow-up), and a minor fail-open/closed doc mismatch. Not gating from me, so it still needs an approver to unblock.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
milosh86: I can't find the doc mentioning it should fail open. It must be run on some older commit as I refactor fail-open to fail-closed intentionally. If an action has a "required" permission, then it makes sense to show it only when validation pass.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
milosh86: hey, yes, I initially implemented `who` checks but it added some complexity I didn't like, so I Ieft it only with `where` check. I would rather leave it as an future improvement, but only if needed, since I don't see currently a case where something else is set as `who` other than DAO. Also, the worst case is that action would be visible, but not executable. Let me know if you see some case where this would make sense to check.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
thekidnamedkd: Ah sorry, not a doc, the PR body here says `filterItemsByPermission` fails open while permissions are loading. If fail-closed is intentional now, code can stay as-is, just wanted to clarify and keep the description aligned.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
thekidnamedkd: The part I’m stuck on is that APP-1001 frames executability as the DAO address being `who`, while this checks `permissionId + where`. If we’re intentionally scoping this to “some permission exists on the target” for now, can we make that explicit in the PR and track `who` as a follow-up? Or maybe I’m not understanding the distinction?

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
milosh86: yeah, I see now in the PR body 👍  ... anyway, fail-closed was the goal.

apps/app/src/modules/governance/components/actionComposer/actionComposerUtils.ts
milosh86: My assumption was that `who` is a DAO in 99.9% of cases in those specific plugin related permissions (which might be a wrong assumption), so it was not worth handling who as  it could be also ANY_ADDRESS.

But yeah, let's better be explicit, I'll add strict DAO expectation for `who`.

milosh86: added who check..

thekidnamedkd (approved):
LGTM.

PR #1255, opened 2026-07-22
feat(APP-1016): support chat on assistant-ui with an agentic backend

tyhonchik:
Rebuilds the support chat as an agentic, tool-calling backend with a stock assistant-ui chat UX themed for Aragon. Obsoletes the deterministic intake pipeline and the hand-rolled widget. Two commits, one per layer.

## Backend — `apps/assistant` + `packages/assistant-contracts`
- One streamed `/chat` turn runs a `streamText` agent with a `createLinearTicket` tool gated behind an explicit user approval (created only after the user approves the draft, deduped by tool-call id, capped per session) plus an auto-approved `flagOffTopic` tool.
- The classify/extract/preview steps and the `/issues` endpoints are gone; refusals live in the system prompt; ticket fields are shared through new `ticket` contract schemas; `maxIssuesPerSession` → 3.

## Widget — `packages/assistant-chat`
- Transcript, streaming and tool cards run through an assistant-ui runtime over our existing chat transport.
- Ticket creation is a `createLinearTicket` approval card: streamed draft → **Create** / **Dismiss** → success link or retryable failure.
- Greeting screen with three intake shortcuts + device-local request history; our own composer (file picker, drag-and-drop, paste, stop) is kept.
- Public `AssistantChat` API preserved, with an optional `getExtendedContext` seam for future DAO-aware context. The old controller state machine, ticket hook and hand-rolled panels are removed.

## Notes
- Retry on a failed creation regenerates the turn — the backend always gates creation behind approval, so retry re-drafts and asks to approve again (not a silent re-fire).
- Widget jest tests render the real runtime in a jsdom environment augmented with Node's WHATWG globals (no polyfill dependency); the ticket approval flow is covered by two integration tests over a faked v7 SSE stream (happy path + failure/retry, monitoring asserted PII-free).
- Live validation (real Linear ticket, cap, dismiss, off-topic, attachments, prompt injection) runs on the stand — local AI Gateway keys are not available here.

Green: root `type-check` (5/5), `lint:check` (4/4), tests assistant 58 / contracts / assistant-chat 2; `validate:changesets` clean (changesets split by release scope — `assistant-chat` is app-scope).


apps/assistant/src/routes/files.ts
```diff
@@ -282,12 +288,39 @@ export const buildFilesRoute = (deps: IAppDependencies) =>
...
+            // and hand back the queued entry so the ticket carries the file once. Best effort:
+            // the queue snapshot predates the claim, so a perfectly concurrent duplicate can
+            // still slip through.
+            const contentHash = await computeContentHash(data);
+            const duplicateFile = queuedFiles.find(
+                (queuedFile) =>
+                    queuedFile.contentHash === contentHash &&
+                    queuedFile.size === validated.size,
+            );
+            if (duplicateFile != null) {
```
thekidnamedkd: Nice that the dedup keeps a ticket from carrying the same file twice. One edge though: it hands back the original queued entry and deletes the new blob, so the new tile aliases the original `fileId`. If the user removes that tile after re-attaching the same screenshot and seeing a dupe, DELETE `/files/:fileId` drops the original from the queue and its blob, while an earlier tile (maybe in an already-sent message) still shows attached, and the ticket ends up without a file they think is there. Could we queue the duplicate as its own entry and dedup by `contentHash` at transfer time, or drop the confirm-time dedup since per-ticket dedup was the real goal?

apps/assistant/src/routes/chat.ts
```diff
@@ -284,4 +275,39 @@ const isUpstreamRateLimited = (error: unknown): boolean =>
 // The request schema validates the exact subset of UIMessage the pipeline reads (text parts and
 // roles); the cast bridges the contracts type to the AI SDK type in this single place.
 const toUiMessages = (messages: IChatMessage[]): UIMessage[] =>
-    messages as unknown as UIMessage[];
+    messages
+        .map(dropReasoningParts)
+        .map(resolveDanglingApprovals) as unknown as UIMessage[];
```
thekidnamedkd, line 347: Really like how much simpler this resume path is. One case I think slips through: `resolveDanglingApprovals` handles `approval-requested`, but the composer's Stop can leave the tool part in `input-streaming`/`input-available` (`cancelPendingToolCallsOnSend` is false), and the client sends it back as-is. `convertToModelMessages` runs here without `ignoreIncompleteToolCalls`, so those states become a tool call with no response or throw (vercel/ai#7718), and since the broken part stays in history the rest of that session keeps failing until a new chat. Could we extend `toUiMessages` to resolve or drop parts stuck in those states too? Might be worth a test alongside the approval-requested one.

packages/assistant-chat/src/transport/chatTransport.api.ts
```diff
@@ -5,5 +5,3 @@ import type { UIMessage } from 'ai';
  * the stream carries the conversation only, ticket state lives behind the explicit preview flow.
```
thekidnamedkd: nit: small doc drift, the explicit preview flow is gone now so this can just describe the single stream.
```suggestion
 * the stream carries the whole conversation, including the createLinearTicket tool and approval parts.
```

packages/assistant-contracts/src/limits.ts
```diff
@@ -5,14 +5,25 @@ export const assistantLimits = {
...
-    // the turn limit is never cut off by tokens first: the turn count is the graceful limiter.
+    // The agent emits the ticket fields (title, description, steps) as tool-call arguments, and
+    // reasoning models burn invisible thinking tokens — BOTH count against this cap (observed:
+    // gemini-2.5-flash-lite spent 1168 thinking tokens on a draft and got clipped at 1200,
+    // returning an empty turn). Roomy enough that thinking + a full draft is never clipped.
+    maxOutputTokens: 4000,
+    // The agent resends the transcript on every turn, so late turns cost the most. Generous
+    // enough that a conversation hitting the turn limit is never cut off by tokens first: the
+    // turn count is the graceful limiter.
     maxTokensPerSession: 60_000,
```
thekidnamedkd, line 16: One thing worth thinking through on the graceful-limiter intent: the approval resume is its own POST /chat, so it hits these gates before streaming. If the session crosses a limit on that resume, we return the fixed limit message, the tool never runs, and the card sticks on `approved === true` with no result (the "Creating your request…" spinner never resolves). The old flow kept creation open at the limit. Maybe let a resume carrying an approved `createLinearTicket` through, or not count it as a turn? Each session gets 20 turns (one turn = one POST /chat), and since approval is a separate round-trip a created ticket burns 2 of them (the draft, then the Create press), so a chat can hit the wall mid-approval a bit sooner than a user expects.

apps/assistant/src/routes/files.ts
tyhonchik: Good catch. Dropped the confirm-time dedup — every upload queues as its own entry now, and the transfer to Linear dedupes by `contentHash` instead, so the ticket still gets the file once.

apps/assistant/src/routes/chat.ts
tyhonchik: Fixed with `ignoreIncompleteToolCalls: true` on `convertToModelMessages` — it drops exactly those two states, and `approval-requested` keeps its superseded-denial path. Test added.

packages/assistant-chat/src/transport/chatTransport.api.ts
tyhonchik: Applied, thanks.

packages/assistant-contracts/src/limits.ts
tyhonchik: Went with the second option: a resume (last message is the assistant's) doesn't count as a turn, so a ticket costs 1 turn and a Dismiss is free. Didn't exempt resumes from the gates — that marker is client-forgeable and the token gate is the last hard stop. Instead a refusal now writes a `tool-output-error` for the pending approved call, so the card fails cleanly instead of spinning forever.

thekidnamedkd (approved):
LGTM. The Stop/resume/file-dedup threads are fixed at `680267c` with targeted tests, and checks are green at that head. You are behind `main`, so rebase when you are ready to merge.

PR #1256, opened 2026-07-22
feat(APP-985): support mixed approving and vetoing bodies within an SPP stage

tyhonchik:
## Context

An SPP stage can hold multiple governance bodies. The contracts and backend already model **approve vs veto as a per-body property** (each body has its own `proposalType`, and a stage carries both `approvalThreshold` and `vetoThreshold`). The app, however, treated it as a stage-wide property (`isVeto(stage) = vetoThreshold > 0`) and stamped one `resultType` onto every body, so it could not represent a stage that **mixes** approving and vetoing bodies — which Alchemix requires.

This makes approve/veto per-body end-to-end: read, evaluate, present, and author.

## Changes

- **Status evaluation** (`sppStageUtils`): `getStageResultCounts` counts approving and vetoing bodies **independently** by each body's `proposalType`; `isVetoReached`/`isApprovalReached` compare against their own threshold. Veto is checked first and overrides approval (mirrors the Miro "SPP Stage Status" flow). The `isActive` predicate keys on "no approval requirement" (`approvalThreshold === 0`) so mixed stages follow the approval path. Stage-level `isVeto` removed; added per-body `isVetoBody`.
- **Voting terminal & body UI**: body content, external vote/report, multi-body summary, and `getBodyResultStatus` derive approve/veto per rendered body. The stage footer shows **both** an approval requirement and a veto condition for a mixed stage. Empty (timelock) stages carry no type.
- **Governance designer**: approve/veto moves to the body (`setupBodyDialog` selector, default Approve); stage settings replace the single type + threshold with adaptive **approval/veto thresholds** shown by body composition (single-type stages look unchanged); the `ProcessStageType` radio is removed. Reverse-mapping restores existing processes (incl. mixed) into the form. The encoder writes per-body `resultType` and both thresholds; early advance is forced off when a stage has any vetoing body.

## Verification

- `type-check`, Biome lint, and the full app unit suite (**2096 tests**) pass, including new mixed-stage tests (veto overrides approval, per-type counting, per-body encoding, early-advance disabled with a veto body).
- **Confirmed against real data**: Giorgi's Sepolia DAO `0x1F0D1ab1ebbBFeeaD2cc5566dBCb2e7403b42499` (SPP One, stage 0) is indexed on the dev backend and returns a mixed stage — `approvalThreshold: 1, vetoThreshold: 1` with bodies `proposalType: 2 (veto)` + `proposalType: 1 (approve)` — exactly the data this PR consumes.


apps/app/src/plugins/sppPlugin/utils/sppTransactionUtils/sppTransactionUtils.ts
```diff
@@ -322,19 +312,24 @@ class SppTransactionUtils {
...
+            (body) => body.proposalType !== SppProposalType.VETO,
+        );
+        const hasVetoingBody = bodies.some(
+            (body) => body.proposalType === SppProposalType.VETO,
+        );
+
+        const approvalThreshold = hasApprovingBody
+            ? settings.approvalThreshold
+            : 0;
+        const vetoThreshold = hasVetoingBody ? settings.vetoThreshold : 0;
```
thekidnamedkd: APP-985 covers existing stages too, and if an approval-only stage round-trips with `vetoThreshold: 0`, then adding a veto body keeps encoding 0 here, so that body can never veto. Same shape exists for `approvalThreshold: 0` with approving bodies. Can we clamp or validate each threshold to at least 1 whenever that body type exists?

apps/app/src/plugins/sppPlugin/components/sppVotingTerminal/components/sppVotingTerminalBodySummaryFooter.tsx
```diff
@@ -26,39 +26,73 @@ export const SppVotingTerminalBodySummaryFooter: React.FC<
...
     const stageStatus = sppStageUtils.getStageStatus(proposal, stage);
     const isAccepted = stageStatus === ProposalStatus.ACCEPTED;
 
-    // Display stage status component if approval is reached and it is not an optimistic stage
-    // or if it is an optimistic stage that is accepted
-    if ((isApprovalReached && !isVeto) || (isVeto && isAccepted)) {
+    // Show the resolved status instead of the requirements once the outcome is
+    // decided: vetoed (a veto can land before approval), approval reached, a
+    // pure veto stage accepted, or a bodyless timelock.
+    const showStageStatus =
```
thekidnamedkd, line 42: non-blocking question: once approval lands in a mixed stage, this hides the veto requirement even though veto bodies can still act until `endDate`. Can we keep the veto line visible while the veto window is live, or confirm the Miro flow wants the condition hidden here?

thekidnamedkd (commented):
Leaving the main behavior notes inline. Two small nits I didn't pin are the old stage-type locale keys and the token/lock-to-vote `safeParse` hardening riding outside APP-985.

milosh86 (approved):
Looks good 👍  ... I had one comment but you addressed it while I was reviewing already :D

PR #1257, opened 2026-07-23
ci: resolve the finalize guard's release scope at the tested SHA

tyhonchik:
The finalize guard ("no new scope changesets at finalize") classifies leftover changesets using the release-scope policy, but resolved it from the merge commit (post-merge `main`). A scope change merged to `main` mid-release then re-classifies changesets that were legitimately out of scope when `changeset version` ran at release start, and the guard fails a release that added nothing new.

That's exactly how [Release 1.34.0's finalize](https://github.com/aragon/app/actions/runs/30007912997/job/89208358351) failed: the branch was cut on Jul 20 under the pre-mapper scoping (so `.changeset/assistant-chat-widget.md` was correctly left unconsumed), `release-scopes.yml` landed on `main` on Jul 21 and moved `@aragon/assistant-chat` into the app scope, and finalize on Jul 23 judged the old changeset by the new policy. The release was completed manually (tag + GitHub Release).

## Changes

- **Read `release-scopes.yml` from the tested SHA** in both finalize workflows (app and assistant): the guard now mirrors the `--ignore` scoping that `changeset version` actually ran with when the release was cut.
- **Keep the merge-commit policy as a fallback only** for release branches cut before the scope mapper landed (they don't ship the file). `releaseScopes.js` and its `node_modules` still run from the merge-commit checkout, so pre-mapper branches keep working.

After this release ships, every new release branch carries the mapper and the fallback becomes dead weight we can eventually drop.

### DIFF
--- a/.github/workflows/app-release-pr-finalize.yml
+++ b/.github/workflows/app-release-pr-finalize.yml
@@ -31,20 +31,30 @@ jobs:
             .github/actions/gh-pr-get-body
             .github/workflows/scripts/slackNotify.js
 
-      # Full checkout + pnpm install at the merge commit (post-merge main), so the scope
-      # resolution below reads the current release-scopes policy — release branches cut before
-      # the scope mapper landed don't ship it.
+      # Full checkout + pnpm install at the merge commit (post-merge main): releaseScopes.js and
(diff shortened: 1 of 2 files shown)


milosh86 (approved):
👍

PR #1258, opened 2026-07-23
feat(APP-999): add Plausible analytics event hooks

thekidnamedkd:
## Summary

Instruments privacy-safe Plausible custom events across the app's core funnels (APP-999). Runtime loading, the proxied endpoint and env gating are owned by the Plausible pipeline integration (#1260 / SREDO-733) — this PR **only adds events**, sent through the shared `analyticsUtils` pipeline.

## Events

- **Wizard** — `wizard_start` / `wizard_step` (with direction) / `wizard_submit` / `wizard_validation_blocked`.
- **Transaction** — `transaction_start` (`attemptKind`) / `transaction_stage` / `transaction_end` / `transaction_failed`, keyed by `flow` + `transactionKind`, idempotent per lifecycle.
- **Actions** — `action_added` / `action_added_batch` with a low-cardinality `actionCategory`.

## Design

- `plausibleAnalyticsUtils` is a **typed event catalog** (event-name union + payload hygiene) that delegates to `analyticsUtils.trackEvent`, coercing values to strings and dropping nullish props — compile-time safety on event names atop the untyped pipeline.
- **Privacy/cardinality:** failures send `errorClass` only (never raw errors); PII is booleanized (`hasEns`, `hasAvatar`, `hasActions`); action types collapse to a small category set.

## Open items

- **Taxonomy sign-off (Evan):** our names are snake_case; the pipeline's seed events (`'Publish DAO Click'`/`'Create DAO Click'`) are Title-Case prose — both coexist on the create-DAO path today.
- **`deposit_address_copied`** is in the catalog but intentionally unwired (no deposit-address UI yet).
- **Smoke test** needs a non-prod Plausible site (SREDO-733).

## Verification

- Affected Jest suites green (18); `tsc` clean for changed files; Biome clean. Rebased onto `main` after #1260.

### DIFF
--- a/apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
+++ b/apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
@@ -15,6 +15,11 @@ import { useDaoChain } from '@/shared/hooks/useDaoChain';
 import { useNetworkSwitch } from '@/shared/hooks/useNetworkSwitch';
 import { monitoringUtils } from '@/shared/utils/monitoringUtils';
 import { buildIntentId } from '@/shared/utils/pendingTransactionManager';
+import {
+    type PlausibleAnalyticsEventName,
+    type PlausibleAnalyticsProps,
+    plausibleAnalyticsUtils,
+} from '@/shared/utils/plausibleAnalyticsUtils';
 import { NetworkSwitchAlert } from '../networkSwitchAlert';
 import {
     type ITransactionStatusStepMetaAddon,
@@ -58,10 +63,10 @@ export const TransactionDialog = <TCustomStepId extends string>(
         onIndexed,
         network = Network.ETHEREUM_MAINNET,
         transactionType,
+        analytics,
         indexingFallbackUrl,
         disableCancel,
     } = props;
-
     const {
         activeStep,
         steps,
@@ -79,11 +84,16 @@ export const TransactionDialog = <TCustomStepId extends string>(
 
     // Make the onSuccess property stable to only trigger it once on transaction success
     const onSuccessRef = useRef(onSuccess);
+    const analyticsRef = useRef(analytics);
+    analyticsRef.current = analytics;
 
     // Guard so onIndexed fires exactly once: react-query polling yields new status-object
     // identities on re-render, so without this the indexed effect would refire.
     const onIndexedFiredRef = useRef(false);
+    const confirmedAnalyticsFiredRef = useRef(false);
+    const terminalAnalyticsFiredRef = useRef(false);
 
+    const failedAnalyticsFiredRef = useRef(false);
     const { address } = useWalletAccount();
     const { buildEntityUrl } = useDaoChain({ network });
     const {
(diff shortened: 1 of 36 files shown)


apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -80,15 +84,42 @@ export const TransactionDialog = <TCustomStepId extends string>(
...
     } = useNetworkSwitch({ network });
 
+    const getTransactionAnalyticsProps = useCallback(
+        () => ({
+            ...analytics,
+            transactionType,
+            network,
+            chainId: requiredChainId,
+        }),
+        [analytics, transactionType, network, requiredChainId],
```
milosh86: `analytics` is passed in as an object literal so `useCallback` here has no effect.

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -80,15 +84,42 @@ export const TransactionDialog = <TCustomStepId extends string>(
...
+            }
+
+            failedAnalyticsFiredRef.current = true;
+            plausibleAnalyticsUtils.track('transaction_failed', {
+                ...getTransactionAnalyticsProps(),
+                step: stepId,
+                errorClass: error instanceof Error ? error.name : 'unknown',
+            });
+        },
+        [getTransactionAnalyticsProps],
```
milosh86: same here, `getTransactionAnalyticsProps` is based on `analytics` which changes on each render

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -80,15 +84,42 @@ export const TransactionDialog = <TCustomStepId extends string>(
...
+                trackTransactionFailure(error, stepId);
                 transactionDialogUtils.monitorTransactionError(error, {
                     stepId,
                     from: address,
                     ...context,
-                }),
-        [address],
+                });
+            },
+        [address, trackTransactionFailure],
```
milosh86, line 161: ... and here chain continues

apps/app/src/shared/components/wizards/wizard/wizardRoot/wizardRoot.tsx
```diff
@@ -73,8 +137,8 @@ export const WizardRoot = <TFormData extends FieldValues = FieldValues>(
     }, [formState, reset]);
 
     const wizardContextValues = useMemo(
```
milosh86, line 156: I can't remember exactly why, but I think `wizardContextValues` needs to be stable, so it's worth checking if `analytics` is passed as object literal here too.

milosh86 (commented):
a couple of comments plus one AI finding which seems valid:

AI finding:

**Transaction events fire for all 28 TransactionDialog consumers, not just the 4 instrumented ones**

plausibleAnalyticsUtils.track('transaction_start' | 'transaction_stage' | 'transaction_end' | 'transaction_failed', …) runs unconditionally in TransactionDialog, and ...analytics spreads undefined to nothing. So tokenApproveTokensDialog, tokenDelegationDialog, all five aragonProfile* dialogs, the gauge-voter dialogs, sppAdvanceStageDialog, capitalDistributorClaimTransactionDialog, etc. now emit the full transaction lifecycle with flow and transactionKind absent.

analyticsEvents.md:222-294 lists both as required props on all four events. They will be missing on the large majority of emitted events, and the dashboards described in the doc will have a big untagged bucket.

Note the asymmetry inside this same PR: the wizard layer is correctly opt-in (wizardRoot.tsx:88 and wizardForm.tsx:27 both bail when analytics == null), the transaction layer is not. Either gate on analytics != null to match the docs and the wizard, or accept the untagged events and say so in the doc — but pick one.

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
thekidnamedkd: Yep, this was the bad part. I moved analytics behind a ref so the callback doesn’t churn when callers pass an inline object.

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
thekidnamedkd: Same fix here: this helper now reads from the ref. The deps are just the transaction context now (`transactionType`, `network`, `chainId`).

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
thekidnamedkd: This was what caused the auto-advance. The transaction effects now depend on the stable tracker, and there’s a regression test covering recreated analytics props.

apps/app/src/shared/components/wizards/wizard/wizardRoot/wizardRoot.tsx
thekidnamedkd: Checked this too. Create-proposal passes nested analytics props inline, so I stabilized the analytics value itself by flow + prop contents and covered that rerender case. I didn’t try to make the whole wizard context stable since `useStepper` already returns a fresh object; this PR only needs the analytics slice stable.

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -65,11 +70,16 @@ export const TransactionDialog = <TCustomStepId extends string>(
 
     // Make the onSuccess property stable to only trigger it once on transaction success
     const onSuccessRef = useRef(onSuccess);
+    const analyticsRef = useRef(analytics);
+    analyticsRef.current = analytics;
```
milosh86, line 88: we don't need this double init, no?

milosh86: ok, I get it, this is to catch any change

milosh86 (dismissed):
👍

milosh86 (approved):

PR #1259, opened 2026-07-23
feat(APP-1022): support voting on objection-stage proposals

harryburger:
# Description

Adds app support for the Alchemix objection stage ([APP-1022](https://linear.app/aragon/issue/APP-1022/app-voting-on-the-objection-stage), subtask of APP-986).

When a proposal has the `isObjection` flag in its settings, the voting terminal keeps all three options visible but only "Object" (the No option relabelled) can be selected — Yes and Abstain are disabled with a help text explaining that only No votes are allowed during the objection phase. The copy changes accordingly ("Do you want to object to this proposal?", the vote dialog says "to object"). The flag lives on the shared `IPluginSettings` since the SPP module needs to read it as well.

Two more fixes that came out of testing the full flow end to end on Sepolia against the sandbox backend:

* Objection stages start with the first stage's tallies carried over, so their approval is reached the moment they open. The stage status logic treated that as final and showed "accepted" while the window was still running, which hid the voting UI completely. Objection stages are now treated like optimistic stages and stay active until the window closes (`sppStageUtils.isObjectionStage`).
* The vote eligibility guard probed `canVote` with a hardcoded Yes option. The objection contract only accepts No, so every user was told they are not eligible. The probe now uses No on objection bodies.

The results breakdown needed no changes here, the backend already seeds `votesByOption` with the carried-over tallies at proposal creation.

Needs the backend PRs aragon/app-backend#1478, aragon/app-backend#1481 and aragon/app-backend#1479 (deployed together to sandbox via aragon/app-backend#1484). Full flow verified on Sepolia: create → vote → advance → object, with the tally correctly moving from Yes to No after the objection. One open point for design: objection votes still show as "No" in the vote list, we might rename that to "Objected" later.

Note for reviewers: `sppStageUtils.getStageStatus` is also touched by #1256 (APP-985) — whichever merges second has a known one-line conflict in the `isOptimisticStage` computation.

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/apps/app/src/plugins/sppPlugin/utils/sppStageUtils/sppStageUtils.test.ts
+++ b/apps/app/src/plugins/sppPlugin/utils/sppStageUtils/sppStageUtils.test.ts
@@ -575,6 +575,73 @@ describe('SppStageUtils', () => {
             );
         });
 
+        it('returns active for an objection stage while the window is open, even though approval is already reached', () => {
+            const now = '2023-01-01T12:00:00.000Z';
+            const startDate = DateTime.fromISO(now).minus({ days: 2 });
+            const minAdvance = DateTime.fromISO(now).minus({ days: 1 });
+            const maxAdvance = DateTime.fromISO(now).plus({ days: 3 });
+            const endDate = DateTime.fromISO(now).plus({ days: 10 });
+
+            const stages = [
+                generateSppStage({ stageIndex: 0 }),
+                generateSppStage({ stageIndex: 1, approvalThreshold: 1 }),
+                generateSppStage({ stageIndex: 2 }),
+            ];
+            const subProposal = generateSppSubProposal({ stageIndex: 1 });
+            subProposal.settings = {
+                ...subProposal.settings,
+                isObjection: true,
+            };
+            const proposal = generateSppProposal({
+                hasActions: true,
+                stageIndex: 1,
+                settings: generateSppPluginSettings({ stages }),
+                subProposals: [subProposal],
+            });
+
+            getStageStartDateSpy.mockReturnValue(startDate);
+            getStageEndDateSpy.mockReturnValue(endDate);
+            getStageMaxAdvanceSpy.mockReturnValue(maxAdvance);
+            getStageMinAdvanceSpy.mockReturnValue(minAdvance);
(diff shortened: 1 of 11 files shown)


thekidnamedkd (commented):
One non-code mismatch: APP-1022 asks for first-stage voters in the votes list, while APP-986 says no merged vote list is required and this PR leaves vote-list naming open. One test nit inline.

apps/app/src/plugins/sppPlugin/utils/sppStageUtils/sppStageUtils.test.ts
```diff
@@ -923,6 +923,33 @@ describe('SppStageUtils', () => {
         });
     });
 
+    describe('isObjectionStage', () => {
+        it('returns true when the stage sub-proposal has the isObjection settings flag', () => {
```
thekidnamedkd, line 1315: nit: we test `isObjectionStage` directly, but the regression was `getStageStatus` returning final while the objection window was still open. Can we add a status test with approval already reached, an open window, and `isObjection` true?

thekidnamedkd (dismissed):
LGTM. Checked the AC question against APP-986 (the parent story) — it explicitly says no merged vote list is required and the objection votes list should show only new objection votes, so the "first-stage voters in the votes list" line in APP-1022 is a stale mismatch, not something to build here. Carried-over tallies come from backend #1481 and the no-only voting is covered here.

Two non-blocking follow-ups: the test nit above (a getStageStatus regression test with approval reached + open window + isObjection true would lock the actual fix), and a rebase — this is DIRTY against main with the known one-line getStageStatus conflict vs #1256.

thekidnamedkd (dismissed):
LGTM. Current head is rebased and green. Two non-blocking nits from the draft remain, but nothing merge-blocking.

thekidnamedkd (approved):
LGTM. Re-checked the delta since the dismissed approval: the rebase onto #1266 is clean, the getStageStatus regression test is in (approval reached + open window + isObjection asserts ACTIVE, non-objection twin asserts ADVANCEABLE), and the override pass-through matches the default flow. Two non-blocking nits, nothing merge-blocking.

PR #1262, opened 2026-07-24
fix(APP-1023): Add stuck-transaction recovery to the transaction dialog

tyhonchik:
# Description

When a wallet-submitted transaction never mines (e.g. stuck in the RPC mempool with an underpriced gas price — the Katana incident, Sentry [APP-NEXT-2CG](https://aragonorg.sentry.io/issues/APP-NEXT-2CG)), the transaction dialog spins forever on the confirm step with both Retry and Cancel disabled, and the persisted record resumes the dialog straight back into the dead spinner on every reopen.

This change adds a degraded-mode branch while keeping the happy path and the one-action-one-transaction invariant untouched:

- After 90 seconds without a receipt the confirm step switches to a `warning` state: Cancel becomes enabled and an alert guides the user to check the transaction in their wallet. Receipt polling never stops — a late receipt (or a wallet speed-up replacement, via viem's replacement detection) still completes the dialog.
- "Retry transaction" is gated behind an explicit confirmation dialog (same pattern as the duplicate-proposal warning) stating that the action executes twice if the original transaction is later confirmed. A reload-resumed dialog with nothing to re-send discards the stale record and restarts the flow from prepare.
- `submittedAt` is persisted on SUBMITTED records so a reopened dialog with a long-stuck transaction warns immediately instead of restarting the 90-second window. Hydration drops records older than 24h and legacy timestamp-less records so dead transactions cannot haunt resume flows and duplicate warnings forever.
- The receipt wait is pinned to the transaction `chainId`, fixing the resumed-dialog-on-another-chain infinite spinner.
- `replacement transaction underpriced` is mapped to an actionable step error label (raise the gas fee in the wallet when retrying) and classified as an expected error in Sentry; the confirm-step timeout is logged as a warning for monitoring.

Linear: [APP-1023](https://linear.app/aragon/issue/APP-1023)

## Type of Change

## Developer Checklist:


apps/app/src/modules/application/dialogs/retryTransactionAlertDialog/retryTransactionAlertDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    const { onRetry } = location.params;
+
+    const { t } = useTranslations();
+    const { close } = useDialogContext();
+
+    const handleRetry = () => {
+        close();
+        onRetry();
+    };
```
thekidnamedkd, line 25: `close()` closes the whole dialog stack, and this alert opens with `stack: true`, so retry unmounts the transaction dialog before `onRetry()` can move it back to APPROVE. We should close only this alert, same as `AragonProfileReleaseAlertDialog`.

```suggestion
    const handleRetry = () => {
        close(location.id);
        onRetry();
    };
```

apps/app/src/modules/application/dialogs/retryTransactionAlertDialog/retryTransactionAlertDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+                </div>
+            </DialogAlert.Content>
+            <DialogAlertFooter
+                actionButton={{
+                    label: t(`${namespace}.action.retry`),
+                    onClick: handleRetry,
+                }}
+                cancelButton={{
+                    label: t(`${namespace}.action.back`),
+                    onClick: () => close(),
```
thekidnamedkd: Same stack issue on the back button: bare `close()` kills the transaction dialog instead of returning to the warning state.

```suggestion
                    onClick: () => close(location.id),
```

apps/app/src/shared/components/transactionDialog/transactionDialog.tsx
```diff
@@ -25,6 +32,12 @@ import { useManagedTransaction } from './useManagedTransaction';
 
 const indexingStepInterval = 1000;
 
+// How long a broadcast transaction may stay unconfirmed before the confirm step surfaces a warning.
+// Generous enough that a slow (but eventually mined) inclusion never false-positives — wallet-priced
+// transactions land within a few blocks even under congestion — while bounding the "stuck in the
+// mempool" spinner. Receipt polling never stops: a late receipt still completes the dialog.
+const confirmStepTimeout = 300_000;
```
tyhonchik:
```suggestion
const confirmStepTimeout = 90_000;
```

apps/app/src/assets/locales/en.json
```diff
@@ -651,6 +651,20 @@
         "description": "Confirm the onchain transaction in your wallet to move your profile records to the new name.",
         "submit": "Confirm in wallet",
         "transactionInfo": "Rename Aragon name"
+      },
+      "retryTransactionAlertDialog": {
+        "title": "Your previous transaction may still succeed",
+        "description": {
+          "1": "Your transaction has not been confirmed yet. If it eventually succeeds, retrying now could result in two transactions.",
+          "2": "Only retry if the transaction if you are sure the transaction has failed or you intentionally want to send it again."
```
thekidnamedkd: nit: duplicated “transaction if” in the warning copy.

```suggestion
          "2": "Only retry if you are sure the transaction has failed or you intentionally want to send it again."
```

apps/app/src/shared/components/transactionDialog/useManagedTransaction.ts
```diff
@@ -87,7 +96,7 @@ export const useManagedTransaction = (
             .exhaustive();
     }, [intentId]);
 
-    const receipt = useWaitForTransactionReceipt({ hash });
+    const receipt = useWaitForTransactionReceipt({ hash, chainId });
```
milosh86: AI finding worth checking:

The warning window is silently capped at 180s!

`useManagedTransaction.ts:99` calls `useWaitForTransactionReceipt({ hash, chainId })` with no timeout. `viem 2.55.2` defaults to timeout = 180_000 and rejects with `WaitForTransactionReceiptTimeoutError`; the app sets `retry: 0` globally (`reactQuery.ts:7`), so there is no retry.

So for a genuinely stuck transaction:

- t=90s → CONFIRM goes warning, AlertCard + "Resend to wallet" appear ✅
- t=180s → query errors → `queryToStepState` returns 'error' → CONFIRM flips warning → error ("Confirmation failed"), the AlertCard vanishes, `isAwaitingConfirmation` goes false so `isConfirmTimedOut` resets, and the `waitTxError` effect (`transactionDialog.tsx:487-493`) ships a Sentry error for every stuck tx.
- A receipt arriving at t>180s never completes the dialog.

That directly contradicts the two comments the design rests on — `transactionDialog.tsx:38` "Receipt polling never stops: a late receipt still completes the dialog" and `:322` "a receipt result arriving later still wins over the timeout". The recovery UI has a 90-second life, then the user lands on a spurious failure — the exact outcome APP-1023 is meant to remove.

Fix: pass timeout: 0 at useManagedTransaction.ts:99 (viem's const timer = timeout ? setTimeout(...) : undefined disables it on falsy; wagmi's filterQueryOptions doesn't strip timeout, so it reaches viem).

milosh86 (commented):
Looks good, only one AI finding which seems valid + typo from KD's comment.

apps/app/src/assets/locales/en.json
tyhonchik: Fixed in 828d7988 — applied the suggestion.

apps/app/src/shared/components/transactionDialog/useManagedTransaction.ts
tyhonchik: Valid finding, thanks — confirmed against viem 2.53.1 sources: `waitForTransactionReceipt` defaults `timeout = 180_000`, and the timer is created with `timeout ? setTimeout(...)`, so a falsy `0` disables it. wagmi passes `timeout` through to the action untouched. Fixed in 828d7988 with `timeout: 0` + a comment, and the hook test now asserts the wait options.

milosh86 (dismissed):
Looks good 👍

milosh86 (approved):

PR #1266, opened 2026-07-27
feat(APP-984): support delegate vote overrides for Alchemix Token Voting

tyhonchik:
# Description

Adds app support for delegate vote overrides in Alchemix Token Voting ([APP-984](https://linear.app/aragon/issue/APP-984/support-delegate-vote-overrides-for-alchemix-token-voting)).

Alchemix runs a fork of TokenVoting (`aragon/alchemix-contracts`) where a token holder whose balance is delegated to someone else can override their delegate's vote in a separate transaction (`overrideVote`), including "in advance" before the delegate has voted. The rest of the Token Voting experience is unchanged.

A key contract detail shapes the UI: each account has a single vote record with a single vote option — `vote()`, `overrideVote()` and `voteAndOverride()` all move the entire position and only differ in which power sources are counted (own voting power vs. balance reclaimed from the delegate). The UI therefore presents one merged control instead of separate vote and override flows (the "contextual" design variant).

Following the direction agreed with Milos, the frontend treats this as Token Voting with DAO-specific behavior rather than a separate plugin:

* New DAO-level slots `GovernanceDaoSlotId.GOVERNANCE_DAO_SUBMIT_VOTE` and `GOVERNANCE_DAO_BUILD_VOTE_DATA` (`modules/governance/constants/moduleDaoSlots.ts`). `TokenSubmitVote` is now a thin host that renders the DAO slot keyed by `dao.id` with the extracted `TokenSubmitVoteDefault` (the previous controls, unchanged) as fallback — no Alchemix code runs for other DAOs. `voteDialogUtils` resolves a DAO-specific build-vote-data function before falling back to the plugin one.
* New `src/daos/alchemix/` folder: `AlchemixSubmitVote` replaces the whole submit-vote area for override-eligible users with one card — delegate row, a single set of vote options, an "Also vote with your voting power" switch (default on, shown when `canVote`) picking `voteAndOverride` vs `overrideVote`, and a submitted state showing the position with a "Vote submitted" checkmark linking to the block explorer plus a single "Change vote" action. Non-eligible/disconnected users (or on read errors) get the default controls. `useAlchemixOverrideStatus` batches `getAccountSnapshot`, `getVoteRecord`, `canOverrideVote`, `canVote`; the position is driven by the on-chain flags (`hasOverridden` / `votedWithDelegatedVp`), not by the indexed votes, since an override also emits a `VoteCast`.
* The votes list skips fully overridden delegate votes (vote option none, zero power) indexed via the optional `voteOverridden` field on `ITokenVote` (aragon/app-backend#1479); partially overridden votes render as default items.
* `TokenVotingOptions` gains a `disabledOptions` prop with a per-option reason, used to disable the user's current option and the option the delegate already voted on; the toggle group's empty deselection value is normalized so an empty option can never be submitted.
* General change: the default Vote button is secondary when the connected user has no voting power.

Only the Sepolia test DAO (`ethereum-sepolia-0x1F0D1ab1ebbBFeeaD2cc5566dBCb2e7403b42499`, deployed by Giorgi) is registered for now; the production DAO id will be added once it exists. The preview environment temporarily points at the sandbox backend (`sandbox.backend.aragonservices.in`) where aragon/app-backend#1484 is deployed — revisit before merge if dev catches up.

Expected conflict: #1259 touches `tokenSubmitVote.tsx` / `tokenVotingOptions.tsx` / `en.json` — whichever merges second resolves (both add option-disabling; straightforward to unify).

## Type of Change

## Developer Checklist:


apps/app/src/plugins/tokenPlugin/components/tokenSubmitVote/components/tokenVotingOptions.tsx
```diff
@@ -72,17 +93,30 @@ export const TokenVotingOptions: React.FC<ITokenVotingOptionsProps> = (
...
+                {voteOptions.map(({ label, value, variant, description }) => {
+                    const disabledOption = disabledOptions?.find(
+                        (option) => option.value === value,
+                    );
+
+                    return (
+                        <TokenVotingOptionToggle
+                            description={
+                                disabledOption?.reason != null
+                                    ? ` ${disabledOption.reason}`
```
thekidnamedkd: nit: this swaps the existing option description for a reason with only a leading space, so the label reads like `Yes Your tokens already count toward Yes`. Can we keep the same separator pattern as the normal option descriptions here? Maybe this is a quick Q for Evan.

tyhonchik: thanks, put `-`
imo it feels better for case "No — Your current override"

apps/app/src/daos/alchemix/utils/alchemixTransactionUtils/alchemixTransactionUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+class AlchemixTransactionUtils {
+    buildVoteData = (
+        params: IBuildVoteDataParams<number, IAlchemixVoteOption>,
+    ): Hex => {
+        const { proposalIndex, vote } = params;
+
+        if (vote.voteType == null) {
+            return tokenTransactionUtils.buildVoteData(params);
```
thekidnamedkd: This DAO-level vote-data slot catches every vote for the Alchemix DAO, but the fallback always encodes Token Voting `vote()` calldata. Fine while the DAO is token-only, but if it gets a multisig, lock-to-vote, or SPP voting body, those votes route here and revert. Can we either fall back to the plugin slot when `vote.voteType` is unset, or document that this DAO slot must handle every plugin on the DAO?

apps/app/src/daos/alchemix/components/alchemixSubmitVoteOverride/alchemixSubmitVoteOverride.tsx
```diff
@@ -0,0 +1,344 @@
...
+                    </p>
+                    <Tag
+                        label={t(
+                            'app.daos.alchemix.alchemixSubmitVoteOverride.delegateTag',
+                        )}
+                        variant="primary"
+                    />
+                </div>
+                <p className="text-neutral-500 text-sm leading-tight">
+                    {delegateeVoteOption != null
```
thekidnamedkd: After a full override drains the delegate’s vote, `getVoteRecord` comes back as `none`, so this header can say the delegate hasn’t voted and still holds the tokens right above the overridden row. Could we branch on `hasOverridden` here so the card doesn’t contradict the override state?

tyhonchik: done

apps/app/src/daos/alchemix/utils/alchemixTransactionUtils/alchemixTransactionUtils.ts
tyhonchik: nice catch

thekidnamedkd (dismissed):
LGTM. Rebased onto current main, prior comments are resolved, and checks are green.

thekidnamedkd (approved):
LGTM. Checks are green and the prior threads are fixed again at current head. Still behind main by one, so it needs a rebase before merge.

thekidnamedkd (changes requested):
I need to pull back my approval here. The `canVote && !canOverride` path opens the change panel for a user with an existing position, but submit stays disabled because it still requires `canOverride`, and this dialog only builds `overrideVote` / `voteAndOverride` payloads. Can we gate the change button on `canOverride` too, or add a real plain-vote path before opening this panel?

thekidnamedkd (approved):
LGTM 🪂

PR #1269, opened 2026-07-29
feat(APP-1030): Add nested actions dialog for composing action lists

milosh86:
# Description

A subtask of APP-987.

Adds a `NestedActionsDialog` to the governance module: an isolated action composer used as the payload editor for actions that take an `Action[]` parameter (e.g. `DAO.execute`, and the upcoming `CrossChainController` forwarding action).

This is the shared groundwork for [APP-1030](https://linear.app/aragon/issue/APP-1030/implement-crosschaincontroller-plugin-basic-action) — the reusable nested-composer piece, without the plugin-specific action itself.

**What's in here**

- **`NestedActionsDialog`** (`modules/governance/dialogs/nestedActionsDialog`) — composes a nested action list in its own `useForm` context. The form is isolated on purpose: the action editor and every custom action component address their fields as `actions.[index]`, so a nested form context lets them be reused as-is while keeping the inner fields out of the outer wizard's validation. On save it validates, runs the actions through `proposalActionPreparationUtils.prepareActions` and hands the prepared list back through `onSubmit`.
- **Decoding of pre-existing actions** — when the dialog is opened to edit a previously composed selection that only carries `to`/`value`/`data`, the raw actions are decoded once via `proposalActionsImportExportUtils.decodeActions`, with `daoId` re-attached (the decoder returns the backend shape, which the basic views need for their own `useDao` lookups). Decode and prepare failures surface as inline alerts and are reported to monitoring.
- **`ProposalActionsEditor`** (`modules/governance/components/proposalActionsEditor`) — extracted verbatim from `CreateExecuteActionsFormActions` so the editor + composer pair can be reused inside the dialog. Adds an `excludeActionTypes` passthrough so an action can be kept out of its own nested composer. `CreateExecuteActionsFormActions` is now a thin wrapper over it.
- **`prepareActions` is generic** over the action type, so callers working with a narrower action shape keep their types through preparation.
- **Dropdown-inside-dialog fixes** — `AutocompleteInputMenu` now uses the GUK dropdown z-index token, sets `pointer-events-auto`, and stops wheel propagation on its scroll container, so the composer's autocomplete behaves correctly when rendered inside a dialog.
- **`stack: true`** on the dialogs opened from within the action composer (wallet-connect action, capital distributor end/pause/resume, gauge registrar/voter actions, multisig remove members) so they stack on top of the composer instead of replacing it — plus the matching note in the `dialog-conventions` rule-skill.

Covered by `nestedActionsDialog.test.tsx`.

## Type of Change

## Developer Checklist:

## Review Checklist:

- [ ] Tested in a preview or locally that all Acceptance Criteria are satisfied


apps/app/src/modules/governance/components/actionComposer/actionComposer/actionComposer.tsx
```diff
@@ -172,7 +172,7 @@ export const ActionComposer: React.FC<IActionComposerProps> = (props) => {
             daoAddress: dao!.address,
             daoNetwork: dao!.network,
         };
-        open(GovernanceDialogId.WALLET_CONNECT_ACTION, { params });
+        open(GovernanceDialogId.WALLET_CONNECT_ACTION, { params, stack: true });
```
thekidnamedkd, line 178: The `stack: true` calls preserve the composer for ESC/outside-click, but most of these child dialogs still self-close with bare `close()`. `dialogProvider` treats that as clearing the whole stack, so confirming or cancelling WalletConnect, capital distributor, gauge voter/registrar, or multisig remove members will also close `NestedActionsDialog` and drop unsaved nested actions once this is wired up. Can we close those children by id before relying on `stack: true`?

milosh86: yeah, good point, I'll need to add missing IDs to close calls.

thekidnamedkd (approved):
LGTM. The stack-close follow-up fixes the open thread: child dialogs close by `location.id`, WalletConnect gets the close handler from the parent, Verify Smart Contract stacks, and the nested dialog closes by its own location id. Checks are green at `6be8358`; you are behind `main`, so rebase when you are ready to merge.

PR #1295, opened 2026-08-04
feat: Implement CrossChainControllerForwardMessageAction basic action (5/10)

milosh86:
# Description

Closes [APP-1029](https://linear.app/aragon/issue/APP-1029).

Implements the `forwardMessage` basic action for the CrossChainController plugin. The plugin, its types, action
definitions and the `GOVERNANCE_PLUGIN_ACTIONS` slot registration landed in APP-1064, but the action component itself was
a placeholder rendering `null` — picking "Forward message" in the action composer produced an empty action body.

The action now renders two inputs and keeps the encoded calldata in sync with them:

- **Destination chain** — a `RadioGroup` built from `settings.crossChain.lanes`, with the DAO's own chain filtered out
  since a controller cannot forward a message to itself. Each card shows the network name and logo resolved from the
  lane's chain id, falling back to `Chain {chainId}` for a chain the app doesn't support. When no lane remains after
  filtering, a warning `AlertInline` explains there is no destination to forward to.
- **Actions** — the batch executed on the destination chain, composed through the `NestedActionsDialog` added in
  APP-1030. Opened with `stack: true` so closing it doesn't tear down the composer, and with
  `excludeActionTypes: [FORWARD_MESSAGE]` so a forward action can't be nested into itself.

On every change the component encodes `forwardMessage(uint256 _destinationChainId, uint256 _gasLimit, bytes _message)`
into the action's `data`, where `_message` is the ABI encoding of the OSx `Action[]` the destination controller hands to
its executor. The three `inputData.parameters` values are written alongside it so the decoded/raw view stays accurate.

Supporting change: `networkUtils.getNetworkByChainId` resolves a standard chain id to a `Network`, which is what lets a
lane be rendered with its network name and logo.

**Known limitation:** the gas limit is hardcoded to 1,000,000 (`TODO(APP-1029)`) until product decides between a user
input and a quote derived from the destination chain. The `repositoryAddresses` in the plugin definition are still zero
addresses from APP-1064, so the plugin is not installable yet — this PR only covers composing the action.

The component asserts `action.daoId != null` (same pattern as `MultisigAddMembersAction`): the nested composer resolves
its DAO data from the action, so this view is DAO-context only.

Dependencies: stacked on `app-1065-properly-handle-decoded-actions-for-unverified-contracts` → `app-1064-…` →
`app-1057-…` → `main`. Merge the bases first.

## Type of Change

## Developer Checklist:

Note for reviewers: `type-check`, `lint` and the existing `networkUtils` / `nestedActionsDialog` suites pass, but the new
component has no unit test of its own yet.

## Review Checklist:


apps/app/src/plugins/crossChainControllerPlugin/components/crossChainControllerActions/crossChainControllerForwardMessageAction/crossChainControllerForwardMessageAction.tsx
```diff
@@ -1,16 +1,311 @@
...
+    >('nestedActions', {
+        label: t(
+            'app.plugins.crossChainController.crossChainControllerForwardMessageAction.actions.label',
+        ),
+        rules: { validate: (value) => value != null && value.length > 0 },
+        fieldPrefix: actionFieldName,
+    });
+
+    const handleDestinationChainChange = (value: string) =>
+        onDestinationChainChange(Number(value));
```
thekidnamedkd: Should changing the destination clear the nested actions? Right now it looks possible to keep the old chain’s batch and re-encode it for the new chain.

apps/app/config/.env.local
```diff
@@ -1,5 +1,5 @@
 # URL of the Aragon backend
-ARAGON_BACKEND_URL=https://dev.backend.aragonservices.in
+ARAGON_BACKEND_URL=https://sandbox.backend.aragonservices.in
```
thekidnamedkd: Was the local backend flip intentional for this PR? If not, can we leave the tracked default on the existing backend?

milosh86: yeah, I'll revert this

apps/app/src/plugins/crossChainControllerPlugin/components/crossChainControllerActions/crossChainControllerForwardMessageAction/crossChainControllerForwardMessageAction.tsx
milosh86: good point, I'll check this!

milosh86: I'll address this in the new top-level PR to avoid conflict resolution

apps/app/config/.env.local
milosh86: I'll add this one too to the new top level PR

thekidnamedkd (dismissed):
LGTM 🚦

thekidnamedkd (dismissed):
LGTM 🚦

harryburger (approved):

PR #1301, opened 2026-08-05
fix(APP-1035): keep uploaded transfer action amount when token data is already cached

harryburger:
# Description

Uploading a transfer action JSON was losing the amount when the asset is not in the DAO. The amount field ended up empty and the decoded view showed the transfer value as 0, so the action user uploads is not the action that would execute.

The cause is a race on mount. When the token details are already cached (which happens exactly in this flow, importing the asset via the composer warms the cache), the effects that re-encode calldata from the form values run in the same pass as the initializer, but still with the placeholder values (amount 0, 18 decimals). They overwrite the amount and the calldata right after the initializer set the real ones, and since rawAmount is cleared there is no retry.

Key changes:

- TransferAssetAction: added an isImportPending flag, the derive/re-encode effects are skipped until the imported action is initialized with the real token decimals. Once initialized they re-run and produce calldata identical to the uploaded file.
- Consolidated the five separate derive effects (data, to, amount, value, inputData) into one effect with a single guard.
- The locked-transfer initializer now only runs for locked transfers, it was also writing the asset field for imported actions.
- Added tests for the component (there were none), covering the cached-token upload, the fresh-fetch upload and the normal composer flow.

Related: [APP-1035](https://linear.app/aragon/issue/APP-1035)

## Type of Change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/.changeset/app-1035-imported-transfer-action-amount.md
+++ b/.changeset/app-1035-imported-transfer-action-amount.md
@@ -1,0 +1,5 @@
+---
+"@aragon/app": patch
+---
+
+Fix uploaded transfer actions losing their amount and calldata when the token details are already cached
--- a/apps/app/src/modules/governance/components/createProposalForm/createProposalFormActions/proposalActions/transferAssetAction/transferAssetAction.test.tsx
+++ b/apps/app/src/modules/governance/components/createProposalForm/createProposalFormActions/proposalActions/transferAssetAction/transferAssetAction.test.tsx
@@ -1,0 +1,231 @@
+import { GukModulesProvider } from '@aragon/gov-ui-kit';
+import { act, render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { FormProvider, useForm, useWatch } from 'react-hook-form';
+import { encodeFunctionData, erc20Abi, parseUnits } from 'viem';
+import * as wagmi from 'wagmi';
+import * as daoService from '@/shared/api/daoService';
+import * as DialogProvider from '@/shared/components/dialogProvider';
+import * as useTokenModule from '@/shared/hooks/useToken/useToken';
+import {
+    generateDao,
+    generateDialogContext,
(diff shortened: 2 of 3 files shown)


milosh86 (commented):
A good one, tested and I couldn't reproduce the bug now. A couple of comments worth checking (some AI, some mine).

Potential minor simplification:
- the useState is redundant. action is actionsMerged (useProposalActionsField.ts:63), i.e. field snapshot ∪ live watched values, so action.rawAmount already goes undefined the instant the init effect clears it. Replacing lines 89–91 with const isImportPending = isImportedErc20Action; and deleting setIsImportPending(false) passes all 3 tests. That drops the useState import, the second source of truth, and the setter.

Bugs that are not introduced by your change, so feel free to open new tickets if it doesn't fit nicely within this PR:

- when action is uploaded, no token logo is rendered
- the gate can never open, and then the UI shows one amount while the calldata sends another. useToken ([shared/hooks/useToken/useToken.ts](https://github.com/aragon/app/blob/fix/app-1035-imported-transfer-amount/shared/hooks/useToken/useToken.ts)) uses allowFailure: false across name/symbol/decimals/totalSupply, so one failing call (RPC hiccup, or a bytes32-name token like MKR) leaves token == null permanently → the init effect never runs → isImportPending never flips → transferAssetAction.tsx:265 is disabled forever. I probed it with isError: true: the user types 7, the field shows "7", and action.data is still the uploaded …498d5880 (1234000000) with _value: "1234000000". transferAssetForm.tsx:75 passes disableMaxValidation={true}, so nothing blocks submitting it. Pre-fix this scenario was also broken (it encoded 7 × 10^18 for a 6-decimal token), so it's an unfixed hole rather than a regression — but the fix makes the mismatch invisible. Minimum: pull isError off useToken and open the gate on it, so it degrades to the old loud-wrong instead of silent-wrong.

milosh86 (approved):
👍

PR #1303, opened 2026-08-05
feat(APP-1004): iterate the support assistant on design feedback

tyhonchik:
This reworks the support assistant panel along Evan's design review. The widget shipped as a straight port of the assistant-ui registry thread: a generic header, a ticket card that led with a "Review your request" label, the request history parked on the greeting screen, and a mail button competing for space in the header. Now the header carries the Aragon mark and a live subline naming what the conversation is working on, the card leads with the ticket itself, and past requests get a view of their own. The spec is the Claude Design project `Aragon Support Assistant.dc.html` plus the divergence list in its `github.md`.

## Changes

- **Header names the object, not only the panel.** Title is `Aragon Assistant`, with a subline underneath: `New conversation` on a fresh chat, `Draft: Clicking any button in the app causes a crash` while the model assembles a draft, `SUP-123: …` once it is filed. It reads the last `createLinearTicket` part in the thread, so it follows the ticket with no extra state. Spent drafts name nothing. A dismissed, superseded or interrupted draft drops the subline back to `New conversation`, the same states the card collapses into a quiet line.
- **Ticket card leads with the ticket.** The `Review your request` heading and the intent tag are gone, the drafted title is the card's heading, the description renders plain with no `Details` micro-label, `Steps to reproduce` is body weight over a numbered list, and the contact reads as a sentence: `We'll reply to evan@aragon.org`. Actions sit behind a divider as `Create ticket` / `Dismiss`.
- **Spent drafts stop being cards.** Superseded, dismissed and interrupted drafts collapse into one centered grey line in the transcript instead of an empty bordered box.
- **Past requests move off the greeting.** A quiet `Past requests (2)` link under the composer opens a list view with a back chevron in the header. The composer is hidden there because it lives inside the thread that the view replaces. Link and list read through one `useRequestHistory` hook. Closing the panel drops the detour, so reopening lands on the conversation.
- **One quiet line under the composer.** On a fresh chat it is the past-requests link; in a conversation it is `Prefer a human? Email support`. The header mail button is dropped, so that escape hatch has a single home.
- **Time dividers.** A divider opens the transcript and marks a resumed conversation (`Today 14:26`, `Monday 09:41`), fired when the pause between two messages passes 30 minutes.
- **Panel owns its own dismissal.** The header close button became a collapse chevron pointing at the panel edge, and `SupportChatTrigger` withdraws from the navigation bar while the panel is open instead of doubling as the close control. Collapsing hands the keyboard back to the trigger that takes the panel's place.
- **Smaller things:** the composer placeholder switches between `Describe the issue…` and `Reply…`, the greeting is now `What do you need help with?`, the user bubble grew a tail, the composer sits on white with a `neutral-200` border, the tool name comes from `createTicketToolName` rather than a literal, the unused `trigger.close` locale key is gone, and the smoke spec follows the renamed header, greeting and `Create ticket` button.

## Note

The design shows a transcript spanning days (`Monday 09:41` next to `Today 14:26`), which cannot happen yet: the session is in-memory, so a reload starts an empty chat. Dividers are therefore gap-based and session-only, and in practice you see exactly one, above the first message. I left transcript persistence out of this PR.

The Aragon mark lives in the widget as its own component, because the package must not import from `apps/app` and gov-ui-kit ships no logo at `^2.9.0`. It is a copy of `aragonLogoIcon.tsx`, worth collapsing if the kit ever gains one.

The collapse chevron is the same at every breakpoint, including the fullscreen drawer below `lg`, where an X would read a little better.


packages/assistant-chat/src/copy.ts
```diff
@@ -19,13 +16,16 @@ const maxFileSizeMb = Math.round(
...
-        title: 'Aragon Support Assistant',
-        close: 'Close',
+        title: 'Aragon Assistant',
+        collapse: 'Collapse chat',
+        back: 'Back to chat',
         startNewChat: 'Start new chat',
+        contextNew: 'New conversation',
+        // Rendered ahead of the drafted ticket title, e.g. "Draft: Voting page crashes".
+        contextDraftPrefix: 'Draft:',
     },
```
thekidnamedkd, line 26: Heads up — the smoke spec still looks for the old support-chat copy, and the preview run has it red even though the job soft-fails. Mind syncing `e2e/tests/smoke/supportChat/supportChat.spec.ts` with this rename?

packages/assistant-chat/src/components/chatHeader/chatHeader.tsx
```diff
@@ -20,47 +28,85 @@ export interface IChatHeaderProps {
...
+
+        if (title == null) {
+            return undefined;
+        }
+
+        const result = ticket.result as ICreateTicketToolOutput | undefined;
+
+        return result != null && ticket.isError !== true
+            ? `${result.identifier}: ${title}`
+            : `${chatCopy.header.contextDraftPrefix} ${title}`;
```
thekidnamedkd: After dismissing a draft, should the header still say `Draft: <title>`? The card says the draft was dismissed, so maybe spent drafts should drop back to `New conversation`.

apps/app/src/modules/application/components/supportChat/supportChatTrigger.tsx
```diff
@@ -5,30 +5,24 @@ import { useFeatureFlags } from '@/shared/components/featureFlagsProvider';
...
 export const SupportChatTrigger: React.FC = () => {
     const { t } = useTranslations();
     const { isEnabled } = useFeatureFlags();
-    const { isOpen, toggle } = useSupportChatContext();
+    const { isOpen, open } = useSupportChatContext();
 
-    if (!isEnabled('supportChat')) {
+    if (!isEnabled('supportChat') || isOpen) {
         return null;
     }
```
thekidnamedkd, line 32: With the trigger hidden while the panel is open, collapsing from the chevron seems to drop keyboard focus to the body. Worth moving focus back to the trigger on close?

packages/assistant-chat/src/components/assistantChat/assistantChat.tsx
```diff
@@ -23,6 +24,10 @@ export const AssistantChat: React.FC<IAssistantChatProps> = (props) => {
 
     const { sessionId, rotate } = useChatSession();
 
+    // The panel shows either the conversation or the requests filed from this device; the composer
+    // lives inside the thread, so swapping the thread out is what hides it in the requests view.
+    const [isViewingRequests, setIsViewingRequests] = useState(false);
```
thekidnamedkd, line 29: If I close the panel from Past requests, should reopening land back there or reset to the chat?

packages/assistant-chat/src/copy.ts
```diff
@@ -72,22 +78,17 @@ export const chatCopy = {
...
-        intentLabel: {
-            feedback: 'Feedback',
-            bug: 'Bug report',
-            support: 'Support request',
-        } satisfies Record<ITicketIntent, string>,
-        descriptionLabel: 'Details',
         stepsLabel: 'Steps to reproduce',
-        contactLabel: 'Contact',
+        // Rendered ahead of the contact the user left, e.g. "We'll reply to evan@aragon.org".
+        contactPrefix: 'We’ll reply to',
```
thekidnamedkd: tiny one: curly apostrophe here vs straight everywhere else in this file — pick one?

thekidnamedkd (commented):
`supportChat.tsx` still mentions the header link for mailing support, but that link moved under the composer here. Mind touching that comment while you’re in this area?

packages/assistant-chat/src/copy.ts
tyhonchik: Synced. The spec now asserts `Aragon Assistant` plus the `New conversation` subline, the new greeting, `Draft: Proposal page crashes` while the card is up, and the `Create ticket` button. The old header mail assertion moved to the `Email support` link under the composer (it only exists once the conversation starts), and the success check reads the ticket link instead of the bare `SUP-123` text, which now also appears in the subline.

packages/assistant-chat/src/components/chatHeader/chatHeader.tsx
tyhonchik: Agreed, fixed. The selector now mirrors the spent states of the card: dismissed or cancelled, superseded (undecided and no longer the last message), interrupted — any of them drops the subline to `New conversation`. A creation that was attempted and failed keeps its title, since that card still offers a retry. Covered by a new dismissal test in `assistantChat.test.tsx`, plus an assertion on the superseded one.

apps/app/src/modules/application/components/supportChat/supportChatTrigger.tsx
tyhonchik: Good catch. The trigger keeps a ref to its button and focuses it when `isOpen` goes back to false — the component itself stays mounted while the panel is open (it only renders null), so the ref is there by the time the effect runs. Test added in `supportChatTrigger.test.tsx`.

packages/assistant-chat/src/components/assistantChat/assistantChat.tsx
tyhonchik: Reset to the chat: the requests are a detour off the conversation, not a place to come back to. `AssistantChat` clears `isViewingRequests` when `isOpen` turns false, so the flip happens during the collapse rather than on the next open. Test added.

packages/assistant-chat/src/copy.ts
tyhonchik: Straightened — every other apostrophe in the file is straight.

thekidnamedkd (approved):
LGTM 💬

PR #1309, opened 2026-08-07
feat(APP-1032): Implement action decoding/details view (9/10)

milosh86:
# Description

Closes [APP-1032](https://linear.app/aragon/issue/APP-1032/implement-action-decodingdetails-view).

Adds the read-only details view for a `forwardMessage` action rendered back from a proposal, using the app's
`actionViewRegistry` (`CrossChainControllerActionType.CROSS_CHAIN_EXECUTE`).

- `CrossChainControllerExecuteActionDetails` shows the destination chain (name + logo, resolved from `chainId` via
  `networkUtils.getNetworkByChainId`, falling back to `Chain {chainId}` for unsupported chains) and the gas limit, then
  re-decodes the `_message` payload's `Action[]` tuple itself (`forwardMessageActionsAbi`, now shared with the create
  view instead of duplicated) to check the backend-decoded sub-actions against.
- `CrossChainControllerNestedActionsList` renders those sub-actions with `ProposalActions.Item` directly — unlike the
  generic `NestedActionsList`, it does *not* resolve a plugin-specific component for them, since the DAO's own network
  and installed plugins belong to the origin chain, not the destination chain the actions execute on.
- `proposalActionUtils.resolveNestedActions` / `buildRawActionStubs` — lifted out of `NestedActionsList` (now just a
  thin caller) so both the generic and cross-chain lists share the same raw-tuple-vs-decoded-length fallback logic.
- `IRawActionTuple` moved to `modules/governance/types` as a shared type instead of being declared locally in
  `NestedActionsList`.

Dependencies: stacked on `app-1054-implement-gaslimit-param` → `app-1056-…` → `app-1053-…` → `main`. Merge the bases
first.

## Type of Change

## Developer Checklist:

## Review Checklist:


apps/app/src/modules/governance/utils/proposalActionUtils/proposalActionUtils.ts
```diff
@@ -25,9 +26,39 @@ import type { IDao, IResource } from '@/shared/api/daoService';
...
+    /**
+     * Resolves the sub-actions of a wrapper action to render, falling back to raw-calldata stubs built from `rawTuple`
+     * when the decoded `subActions` are missing or their length disagrees with the tuple.
+     */
+    resolveNestedActions = (
+        subActions: IProposalAction[] | undefined,
+        rawTuple: IRawActionTuple[],
+    ): IProposalAction[] => {
+        const hasDecodedMismatch =
+            subActions == null || subActions.length !== rawTuple.length;
```
thekidnamedkd: Should we also compare the tuple fields here before trusting `subActions`? Same count can still mean different `to` / `value` / `data`.

apps/app/src/actions/crossChainController/components/crossChainControllerNestedActionsList/crossChainControllerNestedActionsList.tsx
```diff
@@ -0,0 +1,61 @@
...
+    if (actions.length === 0) {
+        return null;
+    }
+
+    return (
+        <ProposalActions.Root actionsCount={actions.length}>
+            <ProposalActions.Container emptyStateDescription="">
+                {actions.map((action, index) => (
+                    <ProposalActions.Item
+                        action={action}
```
thekidnamedkd, line 49: Do we still want the default action normalization here? I get skipping plugin custom views, but the generic defaults seem useful before rendering.

milosh86: we can only show decoded view here ... is there anything else that normalization adds in addition to format update to what basic views expect?

apps/app/src/modules/governance/utils/proposalActionUtils/proposalActionUtils.ts
milosh86: To be very precise, yes, but in practice both rawTuple and actions are coming from the same source (backend decoding API), so this is more a confirmation that all actions are successfully decoded.

On the other side this could be reused in the future so maybe it's better to implement the check properly. I'll add it.

apps/app/src/actions/crossChainController/components/crossChainControllerNestedActionsList/crossChainControllerNestedActionsList.tsx
thekidnamedkd: Yeah, I think the bit worth keeping is the default/basic formatting, not the plugin custom view lookup. Things like transfer amount formatting and metadata rows look pure-data enough to keep the read-only list readable, while still skipping anything that needs origin-chain plugin state. Could we run the default normalization path here, or is there a reason decoded values should stay raw?

milosh86: yeah, you are right, it seems we can keep default normalization ... In this round I wanted to avoid going through all detail action views and checking if it requires DAO or not. That's why I just went with plain decoded views. 

I'll add normalization to top level PR.

thekidnamedkd (dismissed):
LGTM 🪞

thekidnamedkd (dismissed):
LGTM 🪞

harryburger (approved):

PR #1310, opened 2026-08-07
feat: Implement malware scanning

jjavieralv:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List
any dependencies that are required for this change.

## Type of Change

## Developer Checklist:

## Review Checklist:


apps/assistant/src/routes/files.ts
```diff
@@ -288,6 +295,33 @@ export const buildFilesRoute = (deps: IAppDependencies) =>
                 return context.json(body, status);
             }
 
+            // Malware scan of the content itself, after the format gate and before the file is
+            // queued: a flagged file is deleted here and never reaches the queue, the ticket or
+            // the support team. A mandatory engine without a verdict blocks too, but as a
+            // retriable error — see the malwareScan config for the per-engine policy.
+            if (getConfig().malwareScan.enabled) {
```
thekidnamedkd, line 302: Unconfirmed uploads stay public until cleanup. We hand out a public Blob upload first, so someone can stop after `/files/token` + the Blob write and leave an unscanned file around. Could we quarantine until confirm, or scan/delete blobs that never confirm?

apps/assistant/src/files/malwareScanner.ts
```diff
@@ -0,0 +1,176 @@
...
+                    'The scanner returned an unexpected response.',
+                );
+            }
+
+            const { virustotal, claude } = body.verdicts;
+
+            const results = [
+                evaluateEngine({
+                    mode: malwareScan.claude,
+                    status: claude.status,
```
thekidnamedkd, line 163: If `verdicts` exists but `claude` or `virustotal` is missing, this throws before `rejectFile()`. User gets a 500, the claim stays stuck, blob waits for cleanup. Could we validate both nested verdicts first and return `scan_unavailable`?

jjavieralv: good find, but it is even worse because it skips the rejectFile so the retries hit already confirmed. Now its fixed and it return scan_unavailable as you requested. Also we can configure that because we can switch on the config a scan to be optional or mandatory.
I also added test to prevent this from happening in the future.

apps/assistant/src/routes/files.ts
jjavieralv: Those ones are cleaned up on 48h period
Anyway, Im asking Andrii because atm Vercel does not blocking hardly (just IP + other small controls) if someone is uploading files calling it directly. Its from Vercel's design but should be a better way to manage it

tyhonchik:
@jjavieralv please, have a look

apps/assistant/src/routes/files.ts
```diff
@@ -288,18 +303,79 @@ export const buildFilesRoute = (deps: IAppDependencies) =>
                 return context.json(body, status);
             }
 
+            // Rebuild before queueing: the ticket only ever receives bytes the service wrote
+            // itself. A file the sanitizer refuses is dropped like a failed format check.
+            const sanitized = await deps.getFileSanitizer().sanitize(validated);
```
thekidnamedkd: the rebuilt bytes skip the size cap that `validateFile` applied to the upload, and a re-encode can grow a file — sharp's `png()` defaults `palette` off, so an indexed PNG comes back full-colour. when that pushes it over 5 MiB the confirm still returns 201, then `createLinearTicket.ts:49` drops it on re-validation and just logs, so the user sees an attached screenshot the ticket never gets. worth checking `sanitized.size` right here, before the put, so a reject doesn't leave a fresh blob behind?

apps/assistant/src/files/sanitizeFile.ts
```diff
@@ -0,0 +1,152 @@
...
+    const data = await document.save({
+        addDefaultPage: false,
+        updateFieldAppearances: false,
+    });
+
+    return { ...file, data, size: data.byteLength };
+};
+
+export const createFileSanitizer = (): IFileSanitizer => ({
+    sanitize: async (file) => {
```
thekidnamedkd: the ticket asked for a "size and timing check so sanitization fits the upload budget on Vercel functions" and I can't find a bound here. `maxInputPixels` caps decode, and `ParseSpeeds.Fastest` is pdf-lib's yield throttle rather than a deadline, so a 5 MiB PDF with tens of thousands of objects just burns whatever the function allows. is a timeout around `sanitize` in scope, or is that deliberately deferred?

apps/assistant/src/routes/files.ts
```diff
@@ -288,18 +303,79 @@ export const buildFilesRoute = (deps: IAppDependencies) =>
...
+                );
+
+                return context.json(body, status);
+            }
+
+            // Rebuilt bytes go under a fresh blob and the upload is deleted; bytes returned
+            // unchanged (text) keep the uploaded blob. A failed store releases the claim and
+            // keeps the upload, so the confirm can be retried.
+            let storedBlobUrl = blobUrl;
+            if (sanitized.data !== validated.data) {
```
thekidnamedkd: this hinges on `sanitized.data` being a different object, but `IFileSanitizer` is documented as the seam "a further layer (e.g. a self-hosted antivirus) plugs in behind" without saying that returning the same reference means "keep the upload". an implementation that mutates in place and returns the same file would skip the put entirely and queue the original bytes. an explicit `changed` flag on the result would make that unmissable.

apps/assistant/src/files/validateFile.test.ts
```diff
@@ -32,6 +32,16 @@ describe('sanitizeFilename', () => {
...
+    it('strips invisible format characters used to disguise the extension', () => {
+        // A right-to-left override makes the rest of the name render reversed, so this shows up
+        // as "invoiceexe.png" to whoever reads the ticket.
+        expect(sanitizeFilename('invoice\u202Egnp.exe')).toEqual(
+            'invoicegnp.exe',
+        );
+        // Zero-width characters would otherwise let two files carry the same visible name.
+        expect(sanitizeFilename('report\u200B.pdf')).toEqual('report.pdf');
+    });
+
```
thekidnamedkd: the ticket's malicious-sample list names invalid UTF-8, but both text fixtures here carry NUL bytes so they exit early on the `data.includes(0)` check and the strict decoder branch never runs. a lone `0xFF` with no NUL would cover it.

.gitleaksignore
```diff
@@ -37,3 +37,11 @@ f5f578dfd4020c4a38f54eb5980a2310eefe94c3:.yarn/releases/yarn-4.3.0.cjs:generic-a
 59da625af2ff0be48f582d288d73cbf4c52bf105:.yarn/releases/yarn-4.1.1.cjs:generic-api-key:567
 73f7080432457e358b3678fb02d4b5c466e763eb:.yarn/releases/yarn-4.9.1.cjs:generic-api-key:589
 140074ec8d53b2cac8c8bf84f7eecd5522f53201:.yarn/releases/yarn-4.4.0.cjs:generic-api-key:567
+
+# Public ERC-20 contract address used as a test fixture in a variable named `token`
+# (blockchain address, not a credential) — gitleaks reads it as a generic API key. Same
+# finding as the permissionsList entry above, re-fingerprinted after the files moved to
+# apps/app in the monorepo migration.
+8f5ede4daeaaaa56ec7747fbf14a3c527b4c3194:apps/app/src/modules/settings/components/permissionCondition/permissionCondition.test.tsx:generic-api-key:41
```
thekidnamedkd: these three `apps/app` fingerprints look unrelated to attachment sanitization — rebase leftovers from the monorepo move? easier to review separately.

thekidnamedkd (commented):
Worth retitling and filling in the description before this goes further — the title still says "Implement malware scanning" and the body is the untouched template, while the head does the opposite (external scanning removed in favour of in-process sanitization). Same ask as August, and it matters more now that the approach inverted.

apps/assistant/src/routes/files.ts
tyhonchik: Verified, and it grows more than just the palette case: ~1.3x on a realistic indexed PNG, 3 KiB -> 81 KiB on a flat one. So a 4 MiB screenshot can cross the cap.

Check sits right before the put now — 413, claim released, upload blob deleted, no fresh blob left behind. Route test covers it.

apps/assistant/src/files/sanitizeFile.ts
tyhonchik: Measured it instead of guessing. Under the 5 MiB cap: incompressible WebP ~0.9 s, PDF with 60k indirect objects ~0.45 s, flat 48 MP screenshot ~40 ms. Cost tracks entropy, not pixel count.

Skipping the deadline on purpose. `ParseSpeeds.Fastest` is `Infinity`, so the parse never yields and a `Promise.race` cannot fire during it, and libvips work is off-thread and not cancellable. It would shorten the user's wait without freeing the function.

What the ticket asked for is a test now: `upload budget` in sanitizeFile.test.ts, both shapes under a loose 5 s so a shared runner does not flake but an order-of-magnitude regression fails. Numbers are in the comment on `maxInputPixels`.

apps/assistant/src/routes/files.ts
tyhonchik: Fair, that was implicit. The seam returns `{ file, rebuilt }` now, the route branches on the flag, and the interface says an implementation editing bytes in place still reports `true`.

The store-failure test returns the same object with `rebuilt: true`, so the reference comparison cannot come back unnoticed.

apps/assistant/src/files/validateFile.test.ts
tyhonchik: Right, both fixtures died on `includes(0)` before the decoder ever ran. Added a lone `0xFF` inside otherwise valid text, plus a truncated three-byte sequence.

.gitleaksignore
tyhonchik: Not leftovers, though they are unrelated to attachments. `8f5ede4da` is on main, and the lines are fake ERC-20 addresses in `permissionCondition` / `votingPowerConditionSlot` / `permissionsList` fixtures under a variable named `token`, which gitleaks reads as a generic API key. The full-history scan runs on the first push of any new branch, so without them nobody can push.

Can split them out if you prefer, but that PR would have to land first. The durable fix is widening `.gitleaks.toml` so we stop collecting fingerprints per commit — separate devops ticket rather than this branch.

tyhonchik:
Retitled, and the description is filled in. The body was still describing the scanner.

apps/assistant/src/routes/files.ts
```diff
@@ -288,18 +303,96 @@ export const buildFilesRoute = (deps: IAppDependencies) =>
...
+                        'internal',
+                        'The file could not be stored.',
+                    );
+
+                    return context.json(body, status);
+                }
+
+                await deps
+                    .getBlobStore()
+                    .delete([blobUrl])
```
thekidnamedkd: Deleting the upload here leaves a window the rest of the handler doesn't have: `addFile` below is the one call with no try/catch, and if it throws the original is already gone, the rebuilt blob is unreferenced until the cron, and the claim stays held. The retry then takes the `!claimed` branch, finds nothing queued, and answers 400 "The file is already being confirmed." for the next 15 minutes (`sessionStore.ts:11`).

Every other failure path in here releases the claim first. Wrapping the queue write the same way, or deleting after it, would close it.

apps/assistant/src/files/sanitizeFile.ts
```diff
@@ -0,0 +1,170 @@
...
+    'JavaScript',
+    'JS',
+    'Launch',
+    'OpenAction',
+    'AA',
+    'EmbeddedFile',
+    'EmbeddedFiles',
+    'RichMedia',
+    'XFA',
+]);
```
thekidnamedkd: This is exactly the ticket's list, so no complaint, but `/SubmitForm`, `/GoToR` and `/Rendition` survive the rewrite and all three do something on click for whoever opens the ticket — `/SubmitForm` posts form data to a remote URL. Worth adding while the set is right here, or is that deliberately out of scope for p4?

apps/assistant/src/files/sanitizeFile.ts
```diff
@@ -0,0 +1,170 @@
...
+
+const imageEncoders = new Map<string, (image: Sharp) => Sharp>([
+    ['image/png', (image) => image.png()],
+    ['image/jpeg', (image) => image.jpeg({ quality: 90 })],
+    ['image/gif', (image) => image.gif()],
+    ['image/webp', (image) => image.webp()],
+]);
+
+// Decode and re-encode: the output holds pixels only (no metadata, no trailing or foreign
+// chunks). `animated` keeps every frame of a gif/webp; `rotate()` bakes the EXIF orientation
```
thekidnamedkd: nit: `animated` is the one encoder behaviour claimed here that no test covers, and dropping it silently flattens a gif to its first frame. A two-frame fixture asserting the output still has both would pin it.

apps/assistant/README.md
```diff
@@ -23,11 +23,13 @@ All chat/assistant copy is centralized in two places: service-side texts live in
...
+**Sanitization.** Every confirmed upload is rebuilt in-process before it is queued (`src/files/sanitizeFile.ts`, the seam between the magic-byte gate and the session queue in `/files/confirm`): images are decoded and re-encoded with sharp, metadata dropped; PDFs are parsed with pdf-lib (object streams included) and rejected when they carry active content (`/JavaScript`, `/JS`, `/Launch`, `/OpenAction`, `/AA`, `/EmbeddedFile(s)`, `/RichMedia`, `/XFA`) or encryption, then written out afresh; text files are already covered by the strict UTF-8 decode. The rebuilt bytes replace the upload under a fresh blob URL, so the ticket only ever receives files the service wrote itself; a rejected file is deleted and the widget asks the user to export it as an image or a flat PDF. No bytes leave our infrastructure: an external scanner (app#1310 — Cloudflare worker with VirusTotal and Claude) stays off the table until the consent/terms review (APP-1216) is settled; a self-hosted ClamAV behind the same seam is the possible next layer.
```
thekidnamedkd: "the ticket only ever receives files the service wrote itself" isn't true for the text path — those keep the uploaded blob and come back `rebuilt: false` (`sanitizeFile.ts:165`). Validated bytes, just not ours. Same sentence sits in `routes/files.ts:306` if you're editing both.

apps/assistant/src/routes/files.ts
tyhonchik: Right, that was the one path that kept the claim. The upload is deleted after the queue write now, and `addFile` is wrapped like the put: claim released, fresh blob dropped, 500, upload kept for the retry. The over-cap branch deletes both blobs too instead of waiting for the cron. Route test covers the failed write and the retry.

apps/assistant/src/files/sanitizeFile.ts
tyhonchik: Not deliberate, the ticket list was what I started from. Added all three plus `/ImportData`, same class (reads a file on click). `/URI` stays allowed, a plain hyperlink is in most exported docs. Tests for each of the four behind a link annotation, and one that a URI link survives the rewrite.

apps/assistant/src/files/sanitizeFile.ts
tyhonchik: Added a two-frame gif fixture asserting `pages: 2` after the rebuild. Without the flag it comes back as one page, so the test pins exactly that.

apps/assistant/README.md
tyhonchik: Fair, text keeps its upload. Reworded both places: image or PDF reach the ticket only as bytes we wrote, text as uploaded after the strict decode.

PR #1312, opened 2026-08-10
feat(APP-1012): implement support for plugins with unresolved interface type

harryburger:
# Description

Plugins with unresolved interface type were handled inconsistently. They were filtered out only from process lookups, everywhere else they were still returned, so they showed up as tabs and list entries that render nothing.

Now `getDaoPlugins` drops them at the source so every caller gets the same list. Two places opt back in with `includeUnsupported: true`, the permissions view and the contracts version list, because those describe what is actually installed on chain. An unknown plugin can hold EXECUTE_PERMISSION over the DAO, hiding it there would be wrong.

Key changes:

- `daoUtils.isSupportedPlugin` is the single check and `getDaoPlugins` applies it unless `includeUnsupported` is set. Removed the temp PROCESS only filter and its TODO.
- `hasPluginBody` uses the same check, so the members nav link matches what the members page lists. Before it could show the link for a DAO whose only body is unknown, and clicking it just redirects back to dashboard.
- `getAvailablePluginUpdates` now needs the subdomain match and the interfaceType lookup to resolve to the same registered plugin. Before, a plugin with a known subdomain but unresolved interfaceType passed the filter and then crashed the update dialog on `getPlugin(...).installVersion`.
- Updated the `plugin-visibility` rule, there are two filters on the plugin list now and their defaults are opposite.
- Some tests were relying on `generateDaoPlugin()` defaulting interfaceType to unknown, set it explicitly there.

Smoke tested on ethereum-sepolia `0x1770F7e9E22CB5914c5d254093b4174465dB2566`, it has an unknown plugin installed. It is gone from the governance list and the member tabs, still listed under Contracts and still shows its EXECUTE_PERMISSION row in Permissions.

Related: [APP-1012](https://linear.app/aragon/issue/APP-1012)

## Type of Change

## Developer Checklist:

## Review Checklist:

- [x] (For User Stories only) Tested in a preview or locally that all Acceptance Criteria are satisfied


thekidnamedkd (approved):
LGTM 🧩

PR #1318, opened 2026-08-12
ci(APP-1028): group the release summary by Linear tickets

tyhonchik:
Ports the release-summary design shipped in app-backend (`release/0.36.0`, `33cf6073`) to the app's `generateReleaseSummary.js`. Previously the summary emitted one line per commit: a ticket spanning several commits produced several rows, the Linear link only attached to the first one (the `issuesFound` set suppressed repeat links but not repeat rows), and the remaining rows had their scope stripped and read as unrelated work.

## Changes

- **Ticket grouping** — commits are grouped by the Linear tickets they reference (`\b[A-Za-z]{2,}-\d+\b` in the subject always; in the body too for merge-titled squashes). One entry per ticket regardless of commit count, titled by the ticket itself: `[APP-123: Title](url)` plus links to every PR that carried it.
- **Body-harvest cap** — ticket ids are harvested from a merge-titled body only when it references ≤8 tickets, so a squashed bulk-sync PR cannot resurface a whole released changelog.
- **Section ranking** — a ticket lands in the strongest section among its commits: feat/perf > fix > other. Squash bodies (`* feat: …`) classify like subjects.
- **Title fallback** — commits without a resolvable ticket stay visible under their own title (deduplicated); a merge-style title falls back to the first real body line with the PR number appended. This closes the v0.36.0-backend class of incident where a squash that kept the merge subject dropped a feature from the summary.
- **Graceful degradation** — without a Linear token, or on any fetch failure, entries degrade to the title path; the release never fails on Linear. Open (not completed/canceled) tickets still surface in the leading warning section.
- `resolveCommitTitle` is now a pure `(subject, body)` function: bodies come along in one `git log --first-parent --pretty` pass (NUL/record-separated) instead of a `git show` per merge commit.

App specifics are untouched: mainline walk via `--first-parent`, workspace path filters (`readPathFilter`/`commitMatchesPathFilter`), `RELEASE_COMMIT_RE` dropping every workspace's release commits (still checked against the merge-resolved title), GitHub Markdown output.

Companion PR: aragon/github-templates#9 lands the same design in the canonical shared step.

### DIFF
--- a/.github/workflows/scripts/generateReleaseSummary.js
+++ b/.github/workflows/scripts/generateReleaseSummary.js
@@ -68,20 +68,38 @@ const commitMatchesPathFilter = (commit, patterns, git = runGit) => {
 // and would otherwise leak into the summary as "Other Changes".
 const RELEASE_COMMIT_RE = /^Release @aragon\/[^@\s]+@\d+\.\d+\.\d+/;
 
-// A PR landed as a merge commit carries its PR title in the commit body; the subject only says
-// "Merge pull request #N from org/branch". Surface the title so the entry reads like a squash.
-const resolveCommitTitle = (commit, subject, git = runGit) => {
-    const merge = subject.match(/^Merge pull request #(\d+)\b/);
-    if (!merge) {
+// Linear issue identifiers like "APP-1234" embedded in commit messages.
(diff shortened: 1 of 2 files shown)


milosh86 (approved):
👍

PR #1319, opened 2026-08-12
fix(APP-987): Address review feedback on cross-chain controller stack

milosh86:
# Description

Addresses review comments raised across the cross-chain controller PR stack (on top of `app-1055-enable-download-of-nested-actions`).

- **Nested action verification** — `resolveNestedActions` now checks that each decoded sub-action matches the raw `_actions` tuple entry (`to`, `data`, `value`) instead of only comparing array length, so a decoded action naming a different target/value/calldata is never rendered in place of the tuple that actually executes.
- **Exact action value export** — `exportActionsToJSON` now exports `value` as a string instead of `Number(...)`, since wei amounts above `2^53` silently lose precision (or serialize to exponential notation) when narrowed to a JS number.
- **Gas limit fixes** — clears the gas limit field before every recalculation so a stale value from a prior estimate can never look authoritative and submittable after a failed/reverted estimate; rejects fractional input (the masked input accepts the locale radix character even though gas is a whole-number `uint256`).
- **Fee token from plugin settings** — reads the lane's fee token metadata (`token`) from the indexed plugin settings instead of fetching it client-side with `useToken`.
- **Allowed actions lifecycle** — `useAllAllowedActions` now keeps `data` `undefined` while the query is disabled or still fetching (rather than defaulting to `[]`), so callers can tell "not yet known" apart from "no actions allowed"; `NestedActionsDialog` and `ActionComposer` were adjusted to treat the allowlist as "load before render" and surface a loading state instead of a brief empty-composer flash.
- **Clear nested actions on destination change** — changing the destination chain now clears previously composed nested actions, since they target contracts on the old destination chain and cannot survive the switch.

## Type of Change

## Developer Checklist:

## Review Checklist:


thekidnamedkd (commented):
Looks good to me on the code side at `b673bb4f`. The only thing still blocking is the `daoAssets.spec.ts` smoke failure, and that looks pretty unrelated to the cross-chain controller changes. I’d rerun that job once before treating it as a real regression.

thekidnamedkd (approved):

thekidnamedkd (dismissed):
LGTM 🚀

harryburger (approved):

PR #1320, opened 2026-08-13
fix(ci): restore the preview deployment URL and fail loudly when it is lost

tyhonchik:
# Description

Preview deployments have been landing nowhere since 12 Aug: the PR comment reads `🚀 Preview Deployment: [View Here]()`, an empty link that GitHub resolves back to the PR itself, and the E2E job runs against an empty base URL. The deploy itself is fine — its URL never reaches the jobs that consume it.

The dependabot bump in c07a8ec2 repinned `credential-retrieval` to a version whose `alltofile` mode masks every value it writes to `apps/app/.env`. `kv_app_preview` holds `SENTRY_PROJECT`, whose value `app-next` is also the Vercel project name and therefore a substring of every deployment hostname. Masking is job-wide, so `vercel deploy` reported `https://***-8j08cu55t-aragon-app.vercel.app` and the runner dropped the output: `Skip output 'deploymentUrl' since it may contain secret` ([deploy log](https://github.com/aragon/app/actions/runs/31628282442/job/94425319449)). Staging and production alias a fixed domain and pass fixed URLs to E2E, so only previews were affected.

## Changes

- **Repin `credential-retrieval` to 4992848** (aragon/github-templates#10, merged), which leaves values too short to be secrets unmasked and warns naming the vault item instead. Both pins also carry the right commit link again — the bump left the old SHA in the trailing comment.
- **`verify-output` job** — fails the shared deploy when `deploymentUrl` does not reach the workflow output. Callers gate on `needs.deploy.result == 'success'`, so they now skip instead of commenting a dead link or testing nothing.
- **`Set output` reads the last line and rejects an empty URL** — the alias step in the same job would otherwise run `vercel alias` with no argument, and a second stdout line would produce a malformed multi-line output entry.

Verified on this PR's own preview run: both URLs are back in the comment (app `app-next-nujz52s9h`, assistant `assistant-96j9n2ebq`), and `verify-output` passed for the app and the assistant deploy alike.

Follow-up outside CI: `SENTRY_PROJECT` is config, not a credential, and belongs in `pnpm run setup <env>` rather than `kv_app_preview`. The length floor keeps it harmless either way.

## Type of Change

- None (CI-only change, no changeset)

## Developer Checklist:

- [x] Covered by unit tests (`lib/gha.test.js` in the companion PR; this side is workflow YAML)

### DIFF
--- a/.github/workflows/shared-deploy.yml
+++ b/.github/workflows/shared-deploy.yml
@@ -74,7 +74,7 @@ jobs:
       deploymentUrl: ${{ steps.set-deployment-url.outputs.deploymentUrl }}
     steps:
       - name: Download infra credentials
-        uses: aragon/github-templates/steps/credential-retrieval@7eb535980e228a80c9b80b1781362f7da8d41db2 # https://github.com/aragon/github-templates/commit/a36dc232323f483a457ae3ef46028f3a4ef9f2e6
+        uses: aragon/github-templates/steps/credential-retrieval@4992848afd79e5feb5fa4a688674d13e9ed47c22 # https://github.com/aragon/github-templates/commit/4992848afd79e5feb5fa4a688674d13e9ed47c22
         with:
           op-token: "${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}"
           op-vault: "${{ inputs.vault-prefix }}_infra"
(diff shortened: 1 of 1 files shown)


milosh86 (approved):
Looks good 👍

PR #1324, opened 2026-08-19
chore(deps): update dependencies, bump pnpm to 11.21, drop corepack

tyhonchik:
Recurring dependency refresh plus the pnpm/corepack cleanup. `pnpm audit` goes 42 → 2 (prod 31 → 1); dedupe, peers, lint, type-check, tests and next build are all green on the final tree.

## Changes

- **`pnpm up -r` across all workspaces** — Sentry 10.70, viem 2.55.13, wagmi 3.7.6, React 19.2.8, Next 16.3, ai 7.0.60, hono 4.13, turbo 2.10.9, biome 2.5.8. Majors taken: `vercel` 56 → 58.4.4 and `@linear/sdk` 88 → 89 (generated SDK). Deliberately held: typescript 5.9, framer-motion 12, changesets 2, jest-dom 6.9.1, @types/node 24 (matches Node 24), ultracite 7.8.3 (the 7.10 reformat stays a separate PR).
- **New supply-chain holds** — `vercel` pinned EXACT 58.4.4 and `@vercel/oidc` held at 3.8.1: everything newer ships without SLSA provenance and trips the trustPolicy. `@coinbase/cdp-sdk` (transitive of the wagmi connector chain) held at 1.52.0: 1.55.0 lazy-imports optional `@x402/*` peers, which are not installed (`autoInstallPeers: false`) and break `next build`. Each hold carries a drop-condition comment.
- **Fix: clearing the amount field crashed the transfer-asset action** — viem 2.55.13 now throws `InvalidDecimalNumberError` on `''` in `parseUnits`; the `?? '0'` fallback did not catch empty strings.
- **Overrides pruned** — brace-expansion floor widened to `<1.1.18` (closes both advisories), dead assistant-ui cooldown excludes removed. Remaining audit findings are elliptic (low, no upstream patch exists) and deepmerge-ts (high, fix is v8.0.0 which is inside the 7-day cooldown until ~Aug 23 — follow-up bump).
- **`playwright-core` added as an app devDependency** — resolves the long-standing synpress-cache missing-peer warning for real; `pnpm peers check` is clean.
- **pnpm 11.15.1 → 11.21.0, corepack dropped** — pnpm/action-setup reads the `packageManager` field directly and pnpm manages its own version since v10, so the corepack step is gone from the CI setup action, README and design-sync tooling. `ENABLE_EXPERIMENTAL_COREPACK` in shared-deploy.yml is intentionally untouched — that is Vercel's build mechanism and only a real deploy can validate its removal.
- **Lint** — biome 2.5.8 reformat (21 files, formatting only) plus one now-unused suppression removed.


.changeset/deps-update-18-aug-2026-app.md
```diff
@@ -0,0 +1,5 @@
+---
+"@aragon/app": patch
```
thekidnamedkd, line 2: `packages/assistant-chat` takes `ai`, `@ai-sdk/react`, `@radix-ui/react-dialog` and `@vercel/blob` from the catalog as runtime deps, so what ships inside the app bundle moves here, but the widget stays 0.3.0 with nothing in its changelog. It's in the `app` release scope so it can share this file without mixing scopes. Unless we're treating catalog-only bumps as non-events for the libs, in which case say so and I'll stop asking.

```suggestion
"@aragon/app": patch
"@aragon/assistant-chat": patch
```

apps/app/src/modules/governance/components/createProposalForm/createProposalFormActions/proposalActions/transferAssetAction/transferAssetAction.tsx
```diff
@@ -155,7 +155,9 @@ export const TransferAssetAction: React.FC<ITransferAssetActionProps> = (
         ? receiver?.address
         : zeroAddress;
 
-    const weiAmount = parseUnits(amount ?? '0', tokenDecimals);
+    // Fall back on empty string too: viem >= 2.55.13 throws InvalidDecimalNumberError on ''
+    // (e.g. when the user clears the amount field).
+    const weiAmount = parseUnits(amount || '0', tokenDecimals);
```
thekidnamedkd, line 160: nit: the only thing standing on this is the `user.clear()` in "re-encodes the calldata from user edits", which clears and immediately retypes, so the empty-string render is incidental to that test rather than its subject. Leaving the field cleared in one case would turn a future `?? '0'` refactor into a red test instead of a crash in the composer.

pnpm-workspace.yaml
```diff
@@ -119,6 +109,13 @@ peerDependencyRules:
 
 # pnpm overrides
 overrides:
+  # Hold @vercel/oidc at 3.8.1: later releases ship without SLSA provenance and trip the
+  # trustPolicy (no-downgrade). Drop this once provenance returns.
+  '@vercel/oidc@>3.8.1': 3.8.1
+  # Hold @coinbase/cdp-sdk (transitive of @base-org/account) at 1.52.0: 1.55.0 lazy-imports the
+  # optional @x402/* peers, which are not installed (autoInstallPeers: false) and break next build
+  # with "Module not found: @x402/core/client". Drop once the wagmi connector chain handles it.
+  '@coinbase/cdp-sdk@>1.52.0': 1.52.0
```
thekidnamedkd, line 118: Both new holds read well and name their drop-conditions, which is more than most pins get. The one I'd think about is freezing `@coinbase/cdp-sdk` in the wagmi connector chain: it's wallet-adjacent, and a hold that outlives its reason quietly stops taking security patches. Is there anything that makes us revisit these, or does it depend on someone re-reading the comment during the next `pnpm up`?

.github/actions/setup/action.yml
```diff
@@ -40,15 +40,13 @@ runs:
         fetch-depth: ${{ inputs.fetch-depth }}
         token: ${{ inputs.token }}
         ref: ${{ inputs.ref }}
+    # pnpm/action-setup reads the version from the packageManager field in package.json;
+    # pnpm manages its own version since v10, so corepack is not involved anymore.
```
thekidnamedkd, line 44: nit: I scanned the workflows and this cutover is otherwise complete, with `shared-deploy.yml:67` (`ENABLE_EXPERIMENTAL_COREPACK`) the single leftover you already flagged. The deploy job went green on this PR, so is that the validation you wanted, or are we waiting on a production run before dropping it?

.changeset/deps-update-18-aug-2026-app.md
tyhonchik: Fair point, applied your suggestion. A catalog bump that moves a lib's runtime deps is a release event for that lib, so `@aragon/assistant-chat` gets a patch too.

apps/app/src/modules/governance/components/createProposalForm/createProposalFormActions/proposalActions/transferAssetAction/transferAssetAction.tsx
tyhonchik: Added a dedicated test that leaves the field cleared and asserts the zero-transfer encoding, so a future `?? '0'` refactor turns red instead of crashing the composer.

pnpm-workspace.yaml
tyhonchik: It did depend on someone re-reading the comment during the next refresh. I've made that mechanical: the comment now spells out the re-test procedure (drop the override, pnpm install, next build; remove once it passes), and re-checking the overrides block is a standing step of the recurring deps refresh. The pin also only freezes this one transitive, so if a security advisory lands against 1.52.0, pnpm audit flags it in the same refresh regardless of the hold.

.github/actions/setup/action.yml
tyhonchik: The deploy on this PR ran with `ENABLE_EXPERIMENTAL_COREPACK` still set, so it validates the pnpm 11.21 bump through `vercel build` but says nothing about removing the var. Removal is a follow-up PR: its preview deploy exercises the exact same `vercel build` path as production (only `--prod` differs), so a green preview there is the validation, no need to wait for a production run.

thekidnamedkd (approved):
LGTM 🦦

